PR #5, opened 2021-09-21
Redis

tashian:
There's 1-2 changes to default files for systemd here as well.

TODO:
- [ ] Operationalize It still needs prod setup / cert enrollment instructions
- [ ] ACME:
  - [ ] Linux (need to solve the general case—systemd may not be right for this)
  - [ ] Docker (a HOWTO for https://github.com/smallstep/docker-tls/tree/main/)
  - [ ] Kubernetes (@alanchrt)
- [ ] Generic:
  - [ ] Docker
  - [ ] Kubernetes
- [ ] Not sure what else to say about root distribution


src/docs/default/sections/20-operationalize/10-renewal/20-deployments/linux/20-override.mdx
```diff
@@ -0,0 +1,10 @@
+<p>
+    The <inlineCode>cert-renewer@.service</inlineCode> and <inlineCode>cert-renewer@.timer</inlineCode> files define a <i>service unit template</i>.
+    Templates are never run directly,
+    but we will run an instance of them for <DocConfig>{({ name }) => name}</DocConfig> certificate renewals.
+    That service will be called <inlineCode>cert-renewer@<DocConfig>{({ linux }) => linux.systemdUnitName}</DocConfig>.service</inlineCode>,
+    which parallels the <inlineCode><DocConfig>{({ linux }) => linux.systemdUnitName}</DocConfig>.service</inlineCode> unit name
+    used by <DocConfig>{({ name }) => name}</DocConfig> on systemd-based Linux distributions.
+    Because the two units share this value in their name,
+    the <inlineCode>cert-renewer@<DocConfig>{({ linux }) => linux.systemdUnitName}</DocConfig>.service</inlineCode> renewer will attempt to reload or restart <inlineCode><DocConfig>{({ linux }) => linux.systemdUnitName}</DocConfig>.service</inlineCode> after it renews the certificate.
+</p>
```
alanchrt, line 10: This works! Nice default. Will it work for some technologies with no further explanation, or should we drop a note to contribute this section if we don't have it filled in yet?

src/docs/default/sections/20-operationalize/10-renewal/20-deployments/linux/30-enable.mdx
```diff
@@ -0,0 +1,12 @@
...
+<DocConfig>
+{({ linux }) => (
+    <CodeBlock autoCopy language="bash">
+{`sudo systemctl daemon-reload
+sudo systemctl enable --now cert-renewer@${linux.systemdUnitName}.timer`}
+    </CodeBlock>
+)}
+</DocConfig>
+
+You'll see that the timer is active, by checking the output of `systemctl list-timers`.
```
alanchrt, line 12: :+1:

src/docs/nginx.yaml
```diff
@@ -8,6 +8,8 @@ server:
   dnsName: myserver.example.net
   port: 443
 acme: false
+linux:
+  systemdUnitName: nginx
```
alanchrt, line 12: :+1:

src/docs/redis.yaml
```diff
@@ -0,0 +1,23 @@
...
+protocol: RESP
+server:
+  name: redis
+  dnsName: redis.example.net
+  port: 6379
+acme: false
+linux:
+  systemdUnitName: redis-server
+kubernetes:
+  ingressClass: nginx
```
alanchrt, line 14: We can omit for redis. It will default to `nginx`.

src/docs/redis/sections/10-try/20-server.mdx
```diff
@@ -0,0 +1,25 @@
+You can test your certificate by starting up Redis with TLS enabled.
+
+
+<Alert severity="warning">
+  Native "SSL Support" (TLS) was added to Redis 6.0.0, which was released GA on April 30, 2020.
+  TLS in Redis is an <strong>optional feature</strong> and not all Redis server binaries will have it enabled.
+  The official Docker image, and many OS distribution images, have TLS enabled.
```
alanchrt, line 7: I haven't looked at this rendered yet, but sometimes `Alert` content needs a wrapping `<div>` if it contains nested jsx/html components for styles to apply properly. Disregard if it looks fine.

src/docs/redis/sections/10-try/30-test.mdx
```diff
@@ -0,0 +1,3 @@
+Send a `PING` command to Redis to test your TLS configuration:
+
+<MDXBlock path="snippets/test" />
```
alanchrt, line 3: You can also just embed this content directly. I ended up putting it in a snippet for nginx because I wanted to have them test their TLS configuration again after they'd gotten it running in k8s. Then again, maybe we'll want a test step like that after every deplyoment? Should I just duplicate the "test" section down in Operationalize too, at the end?

src/docs/redis/sections/20-operationalize/20-root.mdx
```diff
@@ -0,0 +1,5 @@
+Here are some notes on root distribution for Redis.
+
+- "Adding SAN validation also makes sense (again, optionally) but will probably be more difficult to do, as Redis tends to use IPs rather than hostnames in all sorts of places. [https://github.com/redis/redis/issues/7946#issuecomment-714660881](https://github.com/redis/redis/issues/7946#issuecomment-714660881)
+    - "The best option is to disable matching of hostname to certificate name. As long as your CA is private the security implications should be minor" [https://github.com/redis/redis/issues/7928#issuecomment-714462007](https://github.com/redis/redis/issues/7928#issuecomment-714462007)
+- The `tls-ca-cert-dir` directive is complicated: [https://github.com/redis/redis/issues/7392](https://github.com/redis/redis/issues/7392)
```
alanchrt, line 5: I notice that this is defined as an override for redis. This means that if we have some prose in the default `20-root.mdx` about using your MDM or ansible or whatever, then they won't show for Redis. Intentional, or should we define a second optional section for specifics on a technology?

alanchrt (commented):
Looks good! I think the root distribution notes would be my only concern about merging this WIP.

src/docs/redis.yaml
```diff
@@ -0,0 +1,23 @@
...
+links:
+  - text: 'Redis: TLS Support'
+    url: https://redis.io/topics/encryption
+  - text: 'A Guide to Using Built-in TLS with Redis'
+    url: https://nishanths.svbtle.com/setting-up-redis-with-tls
+  - text: 'Digital Ocean: using stunnel with older versions of Redis that don''t support TLS'
+    url: https://www.digitalocean.com/community/tutorials/how-to-connect-to-managed-redis-over-tls-with-stunnel-and-redis-cli
+  - text: 'Original GitHub issue discussing adding TLS support'
+    url: https://github.com/redis/redis/issues/2178](https://github.com/redis/redis/issues/2178
```
alanchrt, line 23: I just moved links into a `40-further-reading.mdx` file in upstream, so you can delete them here and move to that file. See nginx for an example.

alanchrt:
Merged. Feel free to keep working, just wanted to catch up `main` with your work.

src/docs/default/sections/20-operationalize/10-renewal/20-deployments/linux/20-override.mdx
tashian: It will work by default with some services. For example nginx's service unit has an `ExecReload=` that sends it a HUP signal. So, it needs no override!

src/docs/redis/sections/10-try/30-test.mdx
tashian: hmm I prefer embedding it directly

PR #6, opened 2021-09-21
Use {server.name}-tls format instead of tls-{server.name}

maraino:
### Description

This PR changes the k8s secret name from `tls-{server.name}` to `{server.name}-tls`. The main reason for this is that cert-manager docs adds the suffix `-tls` at the end in their examples. See for example https://cert-manager.io/docs/usage/certificate/#creating-certificate-resources

### DIFF
--- a/src/docs/default/sections/20-operationalize/10-renewal/20-deployments/kubernetes/20-issuer/jwk.mdx
+++ b/src/docs/default/sections/20-operationalize/10-renewal/20-deployments/kubernetes/20-issuer/jwk.mdx
@@ -48,9 +48,9 @@ Create the secret in your cluster, then apply your `StepIssuer` resource.
 
 <CodeBlock
   language="shell"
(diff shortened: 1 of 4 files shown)


alanchrt (approved):

src/docs/default/sections/20-operationalize/10-renewal/20-deployments/kubernetes/20-issuer/jwk.mdx
```diff
@@ -48,9 +48,9 @@ Create the secret in your cluster, then apply your `StepIssuer` resource.
 
 <CodeBlock
   language="shell"
-  copyText=" kubectl create secret generic my-ca-provisioner-password --from-literal=password=Y4nys7f11 && kubectl apply -f my-ca-issuer.yaml"
+  copyText="kubectl create secret generic my-ca-provisioner-password --from-literal=password=Y4nys7f11 && kubectl apply -f my-ca-issuer.yaml"
```
alanchrt, line 51: This was intentional, didn't notice it. Will fix!

PR #32, opened 2021-11-17
Fix systemd renewer script

elyscape:
This corrects the systemctl command. Additionally, it simplifies the shell script used and switches to sh, which unlike bash is guaranteed to be available.

See also smallstep/certificates#748.


tashian:
Fixed in fd31fcb

PR #37, opened 2021-12-15
MySQL PZT

tashian:


src/pzt/default/sections/20-operationalize/20-deployments/kubernetes/30-configuration.mdx
```diff
@@ -3,9 +3,9 @@
...
-  is updated on disk. We haven't yet investigated the specifics for{' '}
-  {props.pzt.name}. Refer to the {props.pzt.name} documentation for more
-  details, and feel free to{' '}
-  <a href="https://github.com/smallstep/docs">contribute this content</a>{' '}
-  directly on GitHub.
+  is updated on disk. A general approach to certificate reloading and secret
+  rotation in Kubernetes is to use{' '}
+  <a href="https://github.com/stakater/Reloader">Reloader</a> or a similar tool
+  to watch the <inlineCode>{props.pzt.server.name}-tls</inlineCode> secret, and
+  run a rolling upgrade of your deployment when it changes.
```
alanchrt, line 10: Hey, that's a great idea!

src/pzt/mysql/sections/10-try/10-certificate.mdx
```diff
@@ -0,0 +1,55 @@
...
+Your certificate and private key will be saved in `server-cert.pem` and `server-key.pem`, respectively.
+
+MySQL is one of a few services that require the intermediate CA certificate to be bundled with the root CA certificate, in order to validate certificate chains. Let's create a bundle of your CA intermediate and root certificates:
+
+```shell
+cat server-cert.pem | awk '/BEGIN/,/END    /{ if(/BEGIN/){a++}; if (a==2) {print} }' > ca.pem
+step ca root >> ca.pem
+```
+
+The certificate bundle will be saved in `ca.pem`.
```
alanchrt, line 57: This looks like a big copy/paste from the default. Is it just the bottom section that changed? If so, I think maybe it makes more sense to use:

```jsx
<MDXBlock default path="sections/10-try/10-certificate" />

... additional MySQL content here ...
```

src/pzt/mysql/sections/10-try/10-certificate.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+Note that MySQL uses RSA keys, thus the `--kty=RSA` here.
+Your certificate and private key will be saved in `server-cert.pem` and `server-key.pem`, respectively.
+
+MySQL is one of a few services that require the intermediate CA certificate to be bundled with the root CA certificate, in order to validate certificate chains. Let's create a bundle of your CA intermediate and root certificates:
+
+```shell
+cat server-cert.pem | awk '/BEGIN/,/END    /{ if(/BEGIN/){a++}; if (a==2) {print} }' > ca.pem
+step ca root >> ca.pem
+```
```
alanchrt: Could we use `<CodeBlock>` with the `copyText` param here so it doesn't copy two lines? Or would this run fine with the newline since it doesn't have the prompts?

src/pzt/mysql/sections/10-try/10-certificate.mdx
alanchrt: Or possibly add the caveat about the `--kty=RSA` flag into `snippets/certificate` instead of `10-certificate`? I pretty much has MySQL and RSA in mind when making that certificates snippet.

src/pzt/mysql/sections/10-try/20-server.mdx
```diff
@@ -0,0 +1,22 @@
+You can test your certificate by starting up MySQL with TLS enabled.
+
+The simplest way to try MySQL with TLS is to use a Docker container for testing.
+
+```shell
```
alanchrt:
```suggestion
```shell nocopy
```

alanchrt: Alternatively, using `<CodeBlock>` with nice `copyText`.

src/pzt/mysql/sections/20-operationalize/20-deployments/kubernetes/10-certificate.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+<p>
+  This configuration specifies that cert-manager should issue and renew a TLS
+  certificate with the DNS name{' '}
+  <inlineCode>{props.pzt.server.dnsName}</inlineCode> and store the certificate
+  and private key in a Kubernetes secret named{' '}
+  <inlineCode>{props.pzt.server.name}-tls</inlineCode>. The certificate is valid
+  for 24 hours, and cert-manager will automatically renew it before expiration
+  and update the <inlineCode>{props.pzt.server.name}-tls</inlineCode> secret.
+</p>
```
alanchrt, line 70: What's the difference from this one and the default? Just the `Certificate` resource configuration? Maybe we should create a snippet for that (eg. `snippets/certificate-resource`) so there's less duplication?

alanchrt (changes requested):
The content on this is absolutely fantastic. I had some comments mostly on copy/paste and content reuse/duplication. Happy to hear your thoughts or to advise there.

Also, I think the unfurl image is the redis one.

src/pzt/mysql/sections/10-try/10-certificate.mdx
tashian: I'll convert it to a CodeBlock with copyText

src/pzt/mysql/sections/10-try/10-certificate.mdx
tashian: unfortunately the snippet is not enough, because I need to use unconventional filenames for the cert and key, and the default has `server.crt` and `server.key` hardcoded. Honestly I don't mind a bit of duplication here, but I think I'm not as much of a stickler about DRYing things up as you are. No right or wrong here though, it's really a matter of preference.

tashian: oh and the `ca.crt` filename is hardcoded too, but I need it to be `ca.pem` (mysql convention)

PR #38, opened 2021-12-28
PostgrSQL PZT

tashian:


src/pzt/postgresql/sections/10-try/10-certificate.mdx
```diff
@@ -0,0 +1,45 @@
...
+  </a>{' '}
+  CLI, bootstrap your CA with <inlineCode>step ca bootstrap</inlineCode> and run
+  the following command (sub the server name for the actual name / DNS name of
+  your {props.pzt.name} server).
+
+</p>
+
+<MDXBlock path="snippets/certificate" />
+
+Your certificate and private key will be saved in `server.crt` and `server.key`, respectively.
```
alanchrt: I think this goes back to the philosophical difference between duplicate vs template. I'm cool with more duplication for now, but maybe we can still keep an eye out and look for ways to DRY it up a bit in the future without losing usability? It's just difficult to tell what is postgres-specific here vs stuff we're already good with on other pages. Admittedly no really constructive ideas on my part how to address it at this point.

alanchrt: Just did a diff locally. It seems like it would still be pretty usable/readable if we made a new snippet in the default `10-certificate`:

```jsx
<MDXBlock path="snippets/root" />
```

Then, instead of duplicating all of `10-certificate`, you just create `snippets/root.mdx` in `postgresql` and make it blank. Not a show-stopper here, but makes PRs more readable without a big blow to composing docs, I think?

src/pzt/postgresql/sections/20-operationalize/30-root.mdx
```diff
@@ -0,0 +1,9 @@
+Once PostgreSQL server TLS is enforced,
+you'll need to make sure that remote clients will verify and trust certificates signed by your CA.
+
+The `psql` CLI client will not read CA certificates from the system trust store
+such as the macOS Keychain.
+It needs to be [explicitly configured](https://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES) to trust your CA,
+and to verify the server's certificate (using `sslmode=verify-full`).
```
alanchrt: This is actually pretty good for calling out `sslmode`. Maybe we should just scream it a little louder? Bold, caps, animated GIF, marquee?

alanchrt (approved):
Content is fantastic. Small DRY comment (to get your opinion) and then a comment on maybe making more noise about the insecure `psql` default.

alanchrt:
Phooey. GitHub somehow lost my first comment about `sslmode`. Re-composing.

src/pzt/postgresql/sections/10-try/30-test.mdx
```diff
@@ -0,0 +1,16 @@
+Now run the `psql` client with strict certificate verification, and check that SSL is in use:
+
+<CodeBlock autoCopy language="shell-session">
+  {`mkdir ~/.postgresql
+step ca root > ~/.postgresql/root.crt
+psql "postgresql://postgres@db.example.net:5432/postgres?sslmode=verify-full"`}
+</CodeBlock>
```
alanchrt, line 7: When I dug in to postgres, it was actually really interesting/surprising to me that `psql` is basically insecure by default:

> By default, PostgreSQL will not perform any verification of the server certificate. This means that it is possible to spoof the server identity (for example by modifying a DNS record or by taking over the server IP address) without the client knowing. In order to prevent spoofing, the client must be able to verify the server's identity via a chain of trust. 

From that [doc you linked](https://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES).

I think it might be worth calling it out more explicitly here (or in research notes) and how important `sslmode` is. It's crazy to me that it's not better by default, and potentially a very surprising fail-open for users.

tashian: Great idea. Added.

PR #48, opened 2022-02-08
Add basic Nebula provisioner info

tashian:
Nebula only has `step beta ca provisioner` support.
But all of our other provisioner docs use `step ca provisioner`.
So, this PR is challenging the question of when do we switch over and get rid of the `step beta ca provisioner` command?
Because I don't want to mix and match docs between `beta` and non-`beta` commands.
@dopey do you have a sense of the timeline for this switchover?


tashian:
I could also say something like "this is an experimental feature that requires the use of `step ca beta` commands and a database-backed provisioner configuration". But we don't have any intro written up about `step ca beta`.

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -33,6 +34,7 @@ Provisioner Capabilities| x509-sign | x509-renew | x509-revoke | ssh-user-cert-s
 JWK    | ✔️  | ✔️  | ✔️  | ✔️  | ✔️  | 𝗫 | 𝗫 | ✔️  | 𝗫
 OIDC   | ✔️  | ✔️  | ✔️  | ✔️  | ✔️ <Reference id="f1" marker="1"/> | 𝗫 | 𝗫 | ✔️  | 𝗫
 X5C    | ✔️  | ✔️  | ✔️  | ✔️  | ✔️  | 𝗫 | 𝗫 | 𝗫 | 𝗫
+Nebula | ✔️  | ✔️  | ✔️  | ✔️  | ✔️  | 𝗫 | 𝗫 | 𝗫 | 𝗫
```
maraino: For SSH we are currently only allowing Host certificates.

maraino (changes requested):

maraino (approved):
LGTM.

Nebula certs have one `Name` and a list of `IPs`. The name is the name of the certificate, it's not a SAN, but it is usually a hostname; and nebula tools only allow setting one IP, more than one might not be supported, but look at all of them.

By default, you can create a certificate where the SANs/Principals match the Name and/or the IPs. And as all the certificate is available in templates you can use generate SANs from any other field if you control the template.

See https://github.com/slackhq/nebula/blob/b5b9d33ee753b510b316f15b503ecb11209fdf9b/cert/cert.go#L36-L45
```golang
type NebulaCertificateDetails struct {
	Name      string
	Ips       []*net.IPNet
	Subnets   []*net.IPNet
	Groups    []string
	NotBefore time.Time
	NotAfter  time.Time
	PublicKey []byte
	IsCA      bool
	Issuer    string


	// Map of groups for faster lookup
	InvertedGroups map[string]struct{}
}
```

PR #49, opened 2022-02-08
Basic info on using TPM2 keys with pkcs#11

tashian:


src/pages/docs/step-ca/configuration.mdx
```diff
@@ -635,23 +635,36 @@ One can only run signing operations using the key.
...
 The `step-pkcs11-init` tool will create or import CA keys and/or certificates onto your HSM, for use with `step-ca`.
 It is configured to work with the YubiHSM2 by default, but you can provide a different PKCS #11 URI by using the `--kms` flag.
 
-Here are some examples of PKCS #11 URIs for various HSMs:
+Here are some examples of PKCS #11 URIs for accessing various devices in Linux:
 
 HSM              | URI format
 ---------------- | -------------------------------------------------------------------------------------
-YubiHSM2 (Linux) | `pkcs11:module-path=/usr/lib/x86_64-linux-gnu/pkcs11/yubihsm_pkcs11.so;token=YubiHSM`
+YubiHSM2         | `pkcs11:module-path=/usr/lib/x86_64-linux-gnu/pkcs11/yubihsm_pkcs11.so;token=YubiHSM`
```
maraino, line 663: YubiHSM2 also needs a pin/password. YubiHSM2 is a bit special because if the pin is `password`, you have to pass 0001password, where 0001 indicates the ID of the authorization key to use.

All of them can use pin-value or pin-source, to pass the password directly or use a file.

PR #56, opened 2022-02-28
Add Fedora support and update CentOS 8 to CentOS Stream 8.

jdoss:
This PR will add support for Fedora 34 and 35 and update CentOS 8 to CentOS Stream 8. CentOS 8 is now EOL.


tashian:
seems legit

PR #58, opened 2022-03-02
Add note on context support in the step-ca command.

tashian:
This is a rare moment where it seems to make sense to mention the step-ca command in the step-cli docs, in order to house all of the docs around contexts in the same place.


tashian:
@dopey just wanted to bump this

dopey (approved):

PR #78, opened 2022-03-21
How to change a JWK provisioner password, and rekeying

tashian:


maraino (approved):
Not sure if all the jq commands are right, but it looks good

PR #79, opened 2022-03-22
Add PostgreSQL database configuration

hslatman:
Support for PostgreSQL was recently added to `step-ca`, but wasn't documented yet. https://github.com/smallstep/certificates/issues/854 was created to track this.

### DIFF
--- a/src/pages/docs/step-ca/configuration.mdx
+++ b/src/pages/docs/step-ca/configuration.mdx
@@ -412,6 +412,40 @@ for supported databases:
   </div>
 </Alert>
 
+*** TLS ***
+
+The PostgreSQL DSN can be used to configure TLS settings for connecting to the server.
+An example DSN that enables TLS server hostname verification is shown below:
+
+```json
+{
+  ...
+  "db": {
+    "type": "postgresql",
+    "dataSource": "postgresql://user:password@127.0.0.1:5432/dbname?sslmode=verify-full",
+  },
+  ...
+},
+```
+
+The database driver will look for a file with root CAs to trust in `$HOME/.postgresql/root.crt`, but will use the system CA trust store if that file is not found.
+Similarly, `$HOME/.postgresql/postgresql.crt` and the corresponding `$HOME/.postgresql/postgresql.key` will be used for mutual TLS authentication if these files exist and if the server requests a client certificate.
+It is also possible to override the locations for these files by providing them in the DSN:
+
+```json
+{
+  ...
+  "db": {
+    "type": "postgresql",
+    "dataSource": "postgresql://user:password@127.0.0.1:5432/dbname?sslmode=verify-full&sslrootcert=/path/to/roots.pem&sslcert=/path/to/client.pem&sslkey=/path/to/client.key",
+  },
+  ...
+},
+```
+
+The behavior of the database driver mimics that of `libpq` and also encompasses handling of other default settings and environment variables.
+More info about the options that `libpq` provides can be found at https://www.postgresql.org/docs/current/libpq-ssl.html.
+
 ### Schema
 
 As the interface is a key-value store, the schema is very simple. We support tables, keys, and values. An entry in the database is a `[]byte value` that is indexed by `[]byte table` and `[]byte key`.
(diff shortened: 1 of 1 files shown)


src/pages/docs/step-ca/configuration.mdx
```diff
@@ -217,11 +218,11 @@ CA's keys.
 
 - **db**: data persistence layer. See [database documentation](#databases) for more info.
 
-  - **type**: badger, bbolt, mysql, etc.
+  - **type**: badger, bbolt, mysql, postgresql, etc.
```
tashian:
```suggestion
  - **type**: badger, bbolt, mysql, or postgresql
```

src/pages/docs/step-ca/configuration.mdx
```diff
@@ -384,6 +386,32 @@ for supported databases:
   </div>
 </Alert>
 
+### PostgreSQL
```
tashian, line 389: Is a server TLS connection to PostgreSQL supported? And if so, can we describe how to write the DSN for it, and where to get the root CA certificate?

My [PostgreSQL PZT](https://smallstep.com/practical-zero-trust/postgresql-tls) might be of some benefit here, or [this part of the libpq-ssl docs](https://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES) (do the same params and defaults apply to Go's psql driver?)

tashian (commented):
added a couple comments

src/pages/docs/step-ca/configuration.mdx
hslatman: Good call. TLS is supported. One example of a DSN is this:

`postgresql://user:password@127.0.0.1:5432/mydb?sslmode=verify-full`

Its behavior is similar to what's in `libpq`. By default it will use the system root CAs as retrieved by Go's `x509.SystemCertPool()` to verify connections to the server. The library will also try default settings for the root CA file and client key / cert pair. It'll look for a directory called `(.)postgresql` in the user home directory to find `root.crt`, `postgresql.key` and `postgresql.crt` (as described here too: https://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES). If these files do not exist, they'll not be used in the configuration. If they are found, the'll be used to configure trusted roots and the client certificate.

It's also possible to explicitly provide a root CA file and/or a private key / cert pair:

`postgresql://user:password@127.0.0.1:5432/mydb?sslmode=verify-full&sslrootcert=/path/to/ca/file.pem&sslcert=/path/to/client.pem&sslkey=/path/to/client.key`

This DSN would result in providing a client certificate at connection setup time, the certificates in the root CA file being loaded as trusted roots and performing server hostname verification (incl. certificate chain verification using the configured trusted roots).

Haven't tried these configurations yet, but I read the code for parsing the DSN, especially the TLS part, at https://github.com/jackc/pgconn/blob/ded272b1f2c31e345b3209d7ab373822bf90b761/config.go#L559-L673.

In general the library mimics the behavior of `libpq` and also supports most of the same parameters. The library also reads parameters from environment variables (e.g. `PGHOST`, `PGUSER`, `PGSSLMODE`, etc; again similar to `libpq`), which are merged with defaults and the DSN. The DSN takes precedence over the environment variables.

What parts of this would you like me to add to the docs? I could also refer to your blog and/or the Postgres docs for more background, of course.

tashian: At minimum, we should let folks know:
- that a TLS connection is possible, and how to enable basic server cert verification with `sslmode=verify-full`
- that it will trust the system CA store
- that it mimics `libpq` behavior, including env vars and default file locations, with a link to their [docs for TLS support](https://www.postgresql.org/docs/current/libpq-ssl.html) for more info.
- and as a bonus, you could show the more complex example of a DSN that explicitly provides a root ca, client cert and client key, and they can pick. I don't expect those parameters will change often, so even though they are for postgres and not step-ca, I think it's OK to include them. (Our docs also have several PKCS#11 URIs for various HSM platforms, for example)

hslatman: I've added a new `TLS` subheader to describe these settings.

tashian: giphy-1

tashian (approved):
Looks great @hslatman! I made a few small tweaks. Go ahead and merge if you feel it's ready

PR #89, opened 2022-04-20
Bump cli and ca versions to 0.19.0 and use latest for step-ssh

dopey:


tashian (approved):
Nice! I tested all the URLs here. Looks good. I also added a note about supported architectures to the SSH host install doc.

PR #91, opened 2022-04-20
Remove language that Arch packages use the Arch User Repository

richardjs:
The `step` and `step-ca` packages are in the official [community](https://wiki.archlinux.org/title/Official_repositories#community) repository now, and thus the Arch User Repository statement can be removed.

The AUR (which is distinct from the community repository) doesn't use `pacman`, and in fact AUR packages require [additional steps](https://wiki.archlinux.org/title/Arch_User_Repository#Installing_and_upgrading_packages) to use. Many Arch users limit their use of the AUR, so it makes sense to update the install docs to reflect the (superior) current situation.

Thanks!


tashian:
Excellent! Thanks @richardjs.

PR #92, opened 2022-04-25
Add Vault RA docs

tashian:
Starting a new PR for the Vault RA documentation.
This is mostly based on @maraino's comment on the previous PR.
I've tested everything locally (Vault v1.10.1, step-ca v0.19.0), and I cleaned up the flow a bit here and there.
I'm still not sure if the policy is too permissive.
If anyone with more Vault expertise can chime in, that would be super helpful!

cc @ahmet2mir @brotherdust @maraino


src/pages/docs/registration-authorities/acme-for-vault.mdx
```diff
@@ -0,0 +1,266 @@
...
+### 1. Start a Vault server and log into vault
+
+For testing, you can run `vault server -dev` in a terminal window, and use an HTTP Vault address as shown here.
+Note: in dev mode, you will lose your PKI when you exit the Vault server.
+
+Now log into Vault:
+
+```bash
+export VAULT_ADDR='http://127.0.0.1:8200'
+vault login token=xxxxx
```
maraino: I believe this is not needed with dev mode.

src/pages/docs/registration-authorities/acme-for-vault.mdx
```diff
@@ -0,0 +1,266 @@
...
+vault write -format=json pki/root/sign-intermediate \
+	csr=@int1.csr \
+	format=pem_bundle \
+	ttl=43800h | jq -r .data.certificate > int1.crt
+cat int1.crt root_ca.crt > intermediate_ca.crt
+vault write pki_int1/intermediate/set-signed \
+	certificate=@intermediate_ca.crt
+vault write pki_int1/config/urls \
+	issuing_certificates="$VAULT_ADDR/v1/pki_int1/ca" \
+	crl_distribution_points="$VAULT_ADDR/v1/pki_int1/crl"
```
maraino: By the way, I was using the names `int1, pki_int1` because you can use a similar approach to do `CA -> INT1 -> INT2`.

src/pages/docs/registration-authorities/acme-for-vault.mdx
```diff
@@ -0,0 +1,266 @@
...
+path "pki_int1/sign/rsa-role" {
+    capabilities = ["create", "update"]
+}
+path "pki_int1/sign/ed25519-role" {
+    capabilities = ["create", "update"]
+}
+path "pki_int/revoke" {
+    capabilities = ["create", "update"]
+}' > policy.hcl
+vault policy write smallstep policy.hcl
```
maraino: According to @brotherdust, this was too permissive.

src/pages/docs/registration-authorities/acme-for-vault.mdx
```diff
@@ -0,0 +1,266 @@
...
+                "certificateAuthorityFingerprint": $vault_ca_fingerprint,
+                "config": {
+                          "pki": "pki_int1",
+                          "pkiRoleRSA": "rsa-role",
+                          "pkiRoleEC": "ec-role",
+                          "pkiRoleEd25519": "ed25519-role",
+                          "appRole": "approle",
+                          "roleID": $vault_role_id,
+                          "secretID": {
+                                "fromString": $vault_secret_id
```
maraino: You can use `fromString`, `fromEnv` or `fromFile`

src/pages/docs/registration-authorities/acme-for-vault.mdx
```diff
@@ -0,0 +1,266 @@
...
+
+Now run: 
+
+```bash
+ROOT_CA_COMMON_NAME="Vault Root CA"
+vault secrets enable pki
+vault secrets tune -max-lease-ttl=87600h pki
+vault write pki/root/generate/internal \
+	common_name="$ROOT_CA_COMMON_NAME" \
+	ttl=87600h
```
maraino: Probably mention that we are creating a 10y old root.

src/pages/docs/registration-authorities/acme-for-vault.mdx
```diff
@@ -0,0 +1,266 @@
...
+INTERMEDIATE_CA_COMMON_NAME="Vault Intermediate CA"
+vault secrets enable -path=pki_int1 pki
+vault secrets tune -max-lease-ttl=43800h pki_int1
+vault write -format=json pki_int1/intermediate/generate/internal \
+	common_name="$INTERMEDIATE_CA_COMMON_NAME" \
+	ttl=43800h  | jq -r .data.csr > int1.csr
+vault write -format=json pki/root/sign-intermediate \
+	csr=@int1.csr \
+	format=pem_bundle \
+	ttl=43800h | jq -r .data.certificate > int1.crt
```
maraino: Probably mention that we are creating a 5y old intermediate.

src/pages/docs/registration-authorities/acme-for-vault.mdx
```diff
@@ -0,0 +1,266 @@
...
+* [`jq`](https://stedolan.github.io/jq/download/) tool to parse JSON output.
+
+When not specified, we run `vault` commands as `root`. A dedicated profile will be created after configuration.
+
+## Configuration Guide
+
+### 1. Start a Vault server and log into vault
+
+For testing, you can run `vault server -dev` in a terminal window, and use an HTTP Vault address as shown here.
+Note: in dev mode, you will lose your PKI when you exit the Vault server.
```
maraino: If the vault server is configured with TLS, we will need TLS we might need to define the CA to use, client certificates, ... This can be achieved with the same environment variables CLI uses, see https://www.vaultproject.io/docs/commands#environment-variables

I haven't tested this.

maraino (commented):
In general, it looks good to me, but we might want to describe a little bit more some of the steps. What the `vault` commands are doing.

src/pages/docs/registration-authorities/acme-for-vault.mdx
maraino: or 
```
Vault CA
     |--- INT1 -- STEP RA1
     |--- INT2 -- STEP RA2
```

src/pages/docs/registration-authorities/acme-for-vault.mdx
brotherdust: I'm sorry I haven't been able to address this yet. I should clarify: it's not that it's too permissive, it's that the policy enables the use of basically all the supported algorithms. The documentation should specify which cipher families to use in which context. For instance, it might not be a good idea to allow RSA on an EC CA, or ed25519 on a CA that will be issuing TLS certs for web services (compatibility reasons). 

Perhaps that is out of the scope of this document; in which case, it would be good to provide a link to reference material that would allow the user to make informed choices about which cipher families to allow via policy. 

Does that make sense? Maybe I'm just overthinking this. Sorry if that's the case!

src/pages/docs/registration-authorities/acme-for-vault.mdx
tashian: Good to know

src/pages/docs/registration-authorities/acme-for-vault.mdx
maraino: Makes sense, but I think the default configuration should allow signing any kind of certificate no matter the key it is using. We can add links to vault docs for more information. 

The same happens with role configurations, if you want to restrict some specific hostnames, you can also do this on vault, but the default role allows everything. You will also be able to manage this in the upcoming policy in `step-ca` instead of doing it on Vault.

tashian: I agree and I will add a sentence to clarify this.

src/pages/docs/registration-authorities/acme-for-vault.mdx
tashian: I've noted this and suggested ppl use `fromFile` in prod

tashian:
Ok @maraino I've resolved the issues you raised here, thanks for the review.
I also broke out the RA documentation into its own docs page.

PR #101, opened 2022-05-23
Docs Search 

tommy-56:
Draft PR for Docs Search. 
Things to do/ could use some guidance
- github actions for algolia gastby plugin


package.json
```diff
@@ -36,11 +40,15 @@
...
     "react-dom": "^16.13.1",
     "react-helmet-async": "^1.1.2",
+    "react-instantsearch-dom": "^6.25.0",
+    "react-instantsearch-hooks-web": "^6.25.0",
     "react-moment": "^0.9.6",
     "refractor": "^2.10.0",
     "rxjs": "^6.5.2",
-    "slugify": "^1.6.0"
+    "slugify": "^1.6.0",
+    "styled-components": "^5.3.5"
```
alanchrt: I think one of the main pieces of work going forward on this will be to eliminate any styled-component dependencies, since we use a different styling solution (material-ui v4 and step-ui). That may be worth starting to dig into... how you will rewrite the included components based on styled-components using material-ui and step-ui. Will work offline to help you get set up on step-ui! Here are the docs for material-ui. Be sure to reference v4 docs, not v5 (we're not upgraded yet):

https://v4.mui.com/

package.json
```diff
@@ -21,13 +21,17 @@
...
+    "algoliasearch": "^4.13.1",
     "cross-fetch": "^3.1.5",
     "formik": "^2.2.9",
     "gatsby": "^3.12.1",
+    "gatsby-plugin-algolia": "^0.26.0",
     "gatsby-plugin-compile-es6-packages": "^2.1.1",
     "gatsby-plugin-mdx": "^2.12.0",
     "gatsby-plugin-next-seo": "^1.8.0",
     "gatsby-plugin-remove-trailing-slashes": "^3.14.0",
+    "gatsby-plugin-styled-components": "^5.14.0",
```
alanchrt: This dependency will end up going away with the port to material-ui and step-ui as well.

src/components/Search/search-result.jsx
```diff
@@ -0,0 +1,58 @@
...
+  return hitCount > 0 ? (
+    <div className="HitCount">
+      {hitCount} result{hitCount !== 1 ? `s` : ``}
+    </div>
+  ) : null;
+});
+
+const PageHit = ({ hit }) => (
+  <div>
+    <Tile>
```
alanchrt: `<ListItem>` (from `@material-ui/core`)

alanchrt: https://v4.mui.com/components/lists/

src/components/Search/search-result.jsx
```diff
@@ -0,0 +1,58 @@
...
+      </Link>
+      <Snippet attribute="excerpt" hit={hit} tagName="mark" />
+    </Tile>
+  </div>
+);
+
+const HitsInIndex = ({ index }) => (
+  <Index indexName={index.name}>
+    <HitCount />
+    <Hits className="Hits" hitComponent={PageHit} />
```
alanchrt:
```javascript

const useStyles = makeStyles(theme => ({
  hits: {
    'ul': {
      listStyle: 'none',
      ...
    }
    'li.ais-Hits-item': {
      marginBottom: '1em',
      'a': {
        color: theme.foreground,
        ....
      }
    }
  }
});

...

const classes = useStyles();

...

<Hits className={classes.hits} ... >

...
```

**Docs:** https://v4.mui.com/styles/basics/#hook-api

src/components/Search/search-box.jsx
```diff
@@ -0,0 +1,28 @@
...
+import Box from '@mui/material/Box';
+
+export default connectSearchBox(
+  ({ refine, currentRefinement, className, onFocus }) => (
+    <form className={className}>
+      <Box
+        component="form"
+        sx= {{
+          width: 350,
+        }}
```
alanchrt:
```suggestion
        width={350}
```

tommy-56:
The styled-components dependencies are gone and its fully using step-ui/mui and is fully functional
Looking for the next round of feedback  👍 @alanchrt

gatsby-config.js
```diff
@@ -1,6 +1,7 @@
-require('dotenv').config({
-  path: `.env.${process.env.NODE_ENV}`,
-});
+// require('dotenv').config({
+//   path: `.env.${process.env.NODE_ENV}`,
+// });
+require("dotenv").config()
```
alanchrt, line 1: I think this should be ok. Do we use `.env` files in any other environments? I'm pretty sure we don't. :+1: 

Could you delete the commented out code?

package.json
```diff
@@ -21,9 +21,13 @@
     "@mdx-js/react": "^1.6.22",
     "@smallstep/internal": "./vendor/internal",
     "@smallstep/step-ui": "./vendor/step-ui",
+    "@styled-icons/fa-solid": "^10.34.0",
```
alanchrt: Is this still needed? We should use all material icons, so could we uninstall this dependency?

package.json
```diff
@@ -21,9 +21,13 @@
...
+    "algoliasearch": "^4.13.1",
     "cross-fetch": "^3.1.5",
+    "eventsource": "1.1.1",
```
alanchrt: Is this specified somewhere as being needed to explicitly install for the plugin? Can we uninstall it?

package.json
```diff
@@ -36,8 +40,11 @@
     "moment-timezone": "^0.5.33",
     "query-string": "^7.0.1",
     "react": "^16.13.1",
+    "react-autosuggest": "^10.1.0",
```
alanchrt: Can we uninstall this one?

package.json
```diff
@@ -36,8 +40,11 @@
...
     "react-dom": "^16.13.1",
     "react-helmet-async": "^1.1.2",
+    "react-instantsearch-dom": "^6.25.0",
+    "react-instantsearch-hooks-web": "^6.25.0",
```
alanchrt: Can we uninstall the `hooks-web` one?

package.json
```diff
@@ -36,8 +40,11 @@
...
     "react-moment": "^0.9.6",
     "refractor": "^2.10.0",
     "rxjs": "^6.5.2",
```
alanchrt, line 46: Could you delete the `package-lock.json` file from the repo? We use `yarn.lock` to lock dependency versions, but package-lock is for npm instead of yarn. :+1:

src/algolia-queries.js
```diff
@@ -0,0 +1,47 @@
...
+      fileAbsolutePath: { regex: "/${escapeStringRegexp(pagePath)}/" },
+    }
+  ) {
+    edges {
+      node {
+        id
+        frontmatter {
+          title
+        }
+        excerpt(pruneLength: 3000)
```
alanchrt: Does this mean we'll only be able to search the first 3k characters of each page rather than the full `body`? I see they have a 10KB limit for index items, which is a bummer. But, maybe we can at least make this a little higher since we don't have much else in the index? 8k?

src/components/Search/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+  ThemeProvider,
+  makeStyles,
+  Box,
+  List,
+  ClickAwayListener,
+} from '@material-ui/core';
+import SearchBox from './search-box';
+import SearchResult from './search-result';
+
+//{window.addEventListener('resize', handleClose)}
```
alanchrt: Delete this one. :+1: (wouldn't work as expected here anyway)

src/components/Search/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+  const [query, setQuery] = useState();
+  const searchClient = useMemo(
+    () =>
+      algoliasearch(
+        process.env.GATSBY_ALGOLIA_APP_ID,
+        process.env.GATSBY_ALGOLIA_SEARCH_KEY
+      ),
+    []
+  );
+  const theme = useStyles();
```
alanchrt: The convention here is to name this `classes`, not `theme`, since it's the generated CSS class names for the styles (eg. in `classes.root`. `theme` is actually a different concept in material-ui, referring to the styling context across all components.

src/components/Search/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+  const [open, setOpen] = React.useState(false);
+  const handleClick = (event) => {
+    setOpen(true);
+  };
+  const handleClose = () => {
+    setOpen(false);
+  };
+  return (
+    <ThemeProvider theme={theme.root}>
```
alanchrt: I think this wrapper shouldn't be needed since we already apply it globally to the whole app. But, similar to what happened above, it looks like you're actually just trying to apply a CSS background. Maybe just make this `<div className={classes.root}?` instead?— after you rename `theme` to `classes`.

src/components/Search/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+          >
+            <SearchBox onInput={handleClick} />
+            {open ? (
+              <Box position={'relative'} zIndex={2} mr={1}>
+                <List
+                  className={theme.popover}
+                  disableAutoFocus={true}
+                  disableEnforceFocus={true}
+                >
+                  <Box border={1} className={theme.resultsBorder}>
```
alanchrt: Is `border={1}` needed? Can this just be a div with `theme.resultsBorder` applied? Also, it would be good to avoid one-off colors. Maybe I can try to get this running locally to see it and make a recommendation.

src/components/Search/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+                    <SearchResult
+                      className={theme.search}
```
alanchrt: Do these styles actually apply successfully when passed this way? I think I'd expect them to live in the `SearchResult` component/file instead. Can we move them there?

src/components/Search/search-result.jsx
```diff
@@ -0,0 +1,106 @@
...
+  const theme = useStyles();
+  return hitCount > 0 ? (
+    <Box borderBottom={1} textAlign={'right'} className={theme.resultsBorder}>
+      <Box mr={2}>
+        {hitCount} result{hitCount !== 1 ? `s` : ``}
+      </Box>
+    </Box>
+  ) : (
+    <Box
+      className={theme.resultsBorder}
```
alanchrt: I'm wondering where we add borders if the `Paper` component would be more appropriate. Will try to get it running and see!

src/components/Search/search-result.jsx
```diff
@@ -0,0 +1,106 @@
...
+const PageHit = ({ hit }) => (
+  <Box mx={-2}>
+    <Link
+      to={
+        hit.title[0] === hit.title[0].toUpperCase()
+          ? `/docs/${hit.slug}`
+          : `/docs/step-cli/reference/${hit.slug}`
+      }
+    >
+      <ListItem button attribute="slug" hit={hit} tagName="mark">
```
alanchrt: Nice you guys got this to work! I think you can delete the `hit` and `tagName` props here.

src/components/Search/search-result.jsx
```diff
@@ -0,0 +1,106 @@
...
+      There are no results found
+    </Box>
+  );
+});
+
+const PageHit = ({ hit }) => (
+  <Box mx={-2}>
+    <Link
+      to={
+        hit.title[0] === hit.title[0].toUpperCase()
```
alanchrt, line 51: Confused. What is this check doing? We're determining URL based on whether the title is all caps?

alanchrt (changes requested):
I left some smaller feedback, but this is really great work! Amazing you guys rolled this altogether as a first substantial React project.

I still haven't run it locally, so I'm going to try to get that going. So, expect a little more feedback on visual tweaks and stuff. But, overall, this is fantastic. Excited to see it ship.

:tada:

package.json
masoncooper19: None of the styled stuff should be needed, we can get rid of that dependency!

src/algolia-queries.js
tommy-56: That is exactly what it is doing.  I believe we had it set much higher about 6k before without issue so we can bump it up a bit 👍

src/components/Search/search-result.jsx
tommy-56: The slugs for the step cli commands are stored like this. The command step beta, for example, the slug is just '/beta' so we had to find a way to append `step-cli/reference/` to the slug. Since the titles of all the cli commands are lowercase this was our way of checking if the search result was a cli command or not.

masoncooper19:
@alanchrt we made some changes to the original tommy-mason doc search branch, we changed things based on your feedback in both pr's. We tried in the past to hide the navbar and just have the list go over but after receiving other feedback from Josh we instead turned our list into a popover essentially. We can try to mess around more with the popover on the other branch you created but I'm having a feeling we're going to be getting a lot more scaling issues that I'm not sure we'd be able to resolve very easily. What do you think?

src/components/Search/search-result.jsx
alanchrt: Ahh, I see. I think they may need to be stored differently if we don't know how to construct the URLs from the data in the index. This will be unreliable and fickle if we end up having a page that has a slug matching this condition unexpectedly.

Is there a way you can just store the full path to individual pages when you store the data in Algolia, then use that path here instead of this workaround?

src/components/Search/index.jsx
alanchrt: This one still needs to be updated to `classes`. :+1:

alanchrt (changes requested):
Recent update looks good! I'd say once the search results are flattened into the side nav (kill the popover), this should be good to merge. Let me know if you need some help on how to hide navigation, since it's in the containing component for search, not nested inside it.

alanchrt:
Example:

search-inline

src/components/Search/search-result.jsx
tommy-56: After playing around today and a lot the other day, I still can't seem to find a way to get a longer version of the slugs into Algolia. I was looking around on the GraphiQL and saw some other graphql queries that would grab the right slug but when I tried using them to talk to Algolia it would throw me an error from Algolia since it did not recognize certain attributes. 
There is an attribute that takes in the full file path but there would still need to be a bit of a work around to trim down all the unnecessary file path info.

I may just be missing some knowledge on Graphql in general or something but for me I could not find a good way to get better slugs into Algolia.

alanchrt: Ok, that stinks, but I don't think it's worth holding up shipping. Once everything else is fully ready for review, let's go ahead and get a last look at it, then our team can come back around and see if we get to fixing this. I made #120 so we don't lose track. :)

tashian:
This PR is my favorite

src/components/Search/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+  resultsOpen: {
+    background: 'white',
+    position: 'relative',
+  },
+  resultsBorder: {
+    border: `1px solid ${classes.palette.divider}`,
+    maxHeight: 580,
+    overflow: 'auto',
+    overflowX: 'hidden',
+  },
```
alanchrt: This border can be deleted now that the results are inline.

```suggestion
```

And then the corresponding `<Box>` can go away too.

src/components/Search/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+        >
+          <SearchBox onInput={handleClick} mt={2} />
+          {openResults ? (
+            <Box position={'relative'} mr={1}>
+              <List
+                className={classes.resultsOpen}
+                disableAutoFocus={true}
+                disableEnforceFocus={true}
+              >
+                <Box className={classes.resultsBorder}>
```
alanchrt: This is the wrapping `<Box>` that can just be deleted.

src/components/Search/search-result.jsx
```diff
@@ -0,0 +1,96 @@
...
+    },
+  },
+}));
+
+const HitCount = connectStateResults(({ searchResults }) => {
+  const hitCount = searchResults && searchResults.nbHits;
+  const classes = useStyles();
+  return hitCount > 0 ? (
+    <Box textAlign={'right'} className={classes.resultsBorder} mb={2} mr={1}>
+      <Typography variant="body1">
```
alanchrt: Smaller size is good here.

```suggestion
      <Typography variant="body2">
```

alanchrt (approved):
@masoncooper19 @tommy-56 I pulled this down and poked around and it looks great! I added a couple very minor style tweaks. Once those are committed, feel free to merge away!

Awesome work. :boom:

PR #105, opened 2022-05-31
Some additions to SCEP documentation

hslatman:
This PR adds a bit of additional documentation for the SCEP provisioner configuration. 

This addresses the concern in https://github.com/smallstep/certificates/issues/746 regarding the key usage expected by the macOS SCEP client. Added a couple of other notes too for completeness, since some more configuration was introduced after the initial release of the SCEP provisioner. Planning on closing that issue after these docs are merged/live.

There'll probably be changes in the SCEP configuration soon if we decide to go forward with the changes discussed internally, but I think we can keep it (largely) backwards compatible with the current implementation and docs.

### DIFF
--- a/src/pages/docs/step-ca/provisioners.mdx
+++ b/src/pages/docs/step-ca/provisioners.mdx
@@ -829,8 +829,8 @@ A shared secret authenticates clients to the CA.
 #### Requirements
 
 Your CA must use an RSA intermediate CA, even if your client supports ECDSA.
-The RSA intermediate is used to decrypt the contents of the SCEP `pkcsPKIEnvelope` and to encrypt the response containing the certificate. 
-These operations cannot be performed using an ECDSA key.
+The RSA intermediate is used to decrypt the contents of the SCEP `pkcsPKIEnvelope` containing the certificate request. 
+This operation cannot be performed using an ECDSA key.
 
 Because [`step ca init`](/docs/step-cli/reference/ca/init) creates an ECDSA chain by default, you will need to [convert your CA to use an RSA CA chain](/docs/tutorials/rsa-chain) before using the SCEP provisioner.
 
@@ -873,7 +873,7 @@ Here's an example of a SCEP provisioner in `$(step path)/config/ca.json`:
 It behaves the same as `forceCN` in the ACME provisioner, and it defaults to false.
 - `challenge` is the secret shared between the provisioner and SCEP clients. By default no secret is used.
 - The `minimumPublicKeyLength` parameter can be used to set the minimum length of public keys submitted by a client. Defaults to 2048.
-- When `includeRoot` is set to true, the root CA certificate will be returned in `GetCACert` requests in addition to the intermediate CA certificate. Defaults to false.
+- When `includeRoot` is set to true, the root CA certificate will be returned in responses to `GetCACert` requests in addition to the intermediate CA certificate. This option was added to support a specific use case for the macOS SCEP client (see [certificates#746](https://github.com/smallstep/certificates/issues/746) for more details). Defaults to false.
 - The `encryptionAlgorithmIdentifier` parameter can be used to change the [encryption algorithm](https://github.com/smallstep/pkcs7/blob/33d05740a3526e382af6395d3513e73d4e66d1cb/encrypt.go#L63) used for encrypting the request content. Defaults to 0: `DES-CBC` for legacy compatibility.
 
 #### Enable the HTTP Server
--- a/src/pages/docs/step-ca/provisioners.mdx
+++ b/src/pages/docs/step-ca/provisioners.mdx
@@ -820,18 +820,31 @@ for more guidance on configuring and using the ACME protocol with `step-ca`.
 
 ### SCEP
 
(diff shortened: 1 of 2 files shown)


src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -820,18 +820,31 @@ for more guidance on configuring and using the ACME protocol with `step-ca`.
...
+SCEP is very popular for use in network equipment and mobile device management (MDM). 
+It runs over HTTP using POSTed binary data or base64-encoded GET parameters, 
+using CMS (PKCS#7) and CSR (PKCS#10) data formats. 
+A shared secret authenticates clients to the CA.
 
 #### Requirements
 
 Your CA must use an RSA intermediate CA, even if your client supports ECDSA.
-We use the RSA intermediate to encrypt/decrypt data, and you cannot do that with an ECDSA key.
+The RSA intermediate is used to encrypt/decrypt data, and you cannot do that with an ECDSA key.
```
tashian: could we elaborate with a couple words what "data" is being encrypted and decrypted with the RSA key?

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -820,18 +820,31 @@ for more guidance on configuring and using the ACME protocol with `step-ca`.
...
+      <strong>Note:</strong> Some SCEP clients may fail if the intermediate CA certificate does not contain the right key usage extensions or does contain otherwise unexpected content. 
+      Consult the documentation of your SCEP client for specific configuration required or ask us on <Link href="https://u.step.sm/discord">Discord</Link> or in <Link href="https://github.com/smallstep/certificates/discussions">GitHub Discussions</Link> if you hit a blocker.
+    </p>
+  </div>
+</Alert>
+
 #### Configure the Provisioner
 
-Add a SCEP provisoiner:
+Add a SCEP provisioner using challenge secret `secret1234` and `AES-256-CBC` as the <Link href="https://github.com/smallstep/pkcs7/blob/33d05740a3526e382af6395d3513e73d4e66d1cb/encrypt.go#L63">encryption algorithm</Link>:
```
tashian:
```suggestion
In this example, we will add a SCEP provisioner using challenge secret `secret1234` and `AES-256-CBC` as the [encryption algorithm](https://github.com/smallstep/pkcs7/blob/33d05740a3526e382af6395d3513e73d4e66d1cb/encrypt.go#L63):
```

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -848,15 +861,25 @@ Here's an example of a SCEP provisioner in `$(step path)/config/ca.json`:
...
+    "encryptionAlgorithmIdentifier": 2,
 }
 ```
 
-- The `forceCN` parameter is optional. It behaves the same as `forceCN` in the ACME provisioner, and it defaults to false.
+- The `forceCN` parameter is optional. 
+It behaves the same as `forceCN` in the ACME provisioner, and it defaults to false.
+- `challenge` is the secret shared between the provisioner and SCEP clients. By default no secret is used.
+- The `minimumPublicKeyLength` parameter can be used to set the minimum length of public keys submitted by a client. Defaults to 2048.
+- When `includeRoot` is set to true, the root CA certificate will be returned in `GetCACert` requests in addition to the intermediate CA certificate. Defaults to false.
```
tashian: Was there a particular client that motivated this option?

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -848,15 +861,25 @@ Here's an example of a SCEP provisioner in `$(step path)/config/ca.json`:
...
+- The `encryptionAlgorithmIdentifier` parameter can be used to change the <Link href="https://github.com/smallstep/pkcs7/blob/33d05740a3526e382af6395d3513e73d4e66d1cb/encrypt.go#L63">encryption algorithm</Link> used for encrypting the request content. Defaults to 0: `DES-CBC` for legacy compatibility.
```
tashian:
```suggestion
- The `encryptionAlgorithmIdentifier` parameter can be used to change the [encryption algorithm](https://github.com/smallstep/pkcs7/blob/33d05740a3526e382af6395d3513e73d4e66d1cb/encrypt.go#L63) used for encrypting the request content. Defaults to 0: `DES-CBC` for legacy compatibility.
```

tashian (approved):
lgtm with some small comments

src/pages/docs/step-ca/provisioners.mdx
hslatman: Added the `pkcsPKIEnvelope` SCEP construct and the response data in b5ad71b12ef1327f1caaab8e4030554e5489fded.

src/pages/docs/step-ca/provisioners.mdx
hslatman: The macOS client. There's some notes about this in https://github.com/smallstep/certificates/issues/746. 

It's been quite a while since then, but I remember there were two cases. One of them required the root to be included in addition to the intermediate, otherwise the request would fail. The other case would happily continue without. They were related to whether the macOS configuration profile had the chain included or not, to function as a trust anchor. So far I don't know about other clients that required this. The macOS client may or may not be doing the right thing, because SCEP's known to have some interoperability issues related to trust anchors. At least this configuration option allows both cases to work.

If we start the implementation of the changes discussed internally, that might affect the macOS client and/or this setting too. I'll have to remain aware of this to ensure compatibility.

tashian: Since it's such a specific setting, seems like it'd be worth mentioning that it's a useful for macOS SCEP clients. Could even link to the issue (I've done that in a couple other places in the docs).

src/pages/docs/step-ca/provisioners.mdx
tashian: 👍🏼

PR #106, opened 2022-05-31
Add policy evaluation documentation

hslatman:
Add docs for certificate issuance policy

High level policy description, rules per name type, some example rules, configuration using `ca.json` and CLI and a couple of CTAs for more granular control in CM/SSH.

P.S.: also did a little trial with creating a `wasm` build of the policy engine (imported as a library) and that seems to work out of the box. That would open up dynamic docs or an interactive form where people can try a policy more easily, while still using the exact same logic as performed by the CA! 😃

### DIFF
--- a/src/pages/docs/step-ca/configuration.mdx
+++ b/src/pages/docs/step-ca/configuration.mdx
@@ -24,7 +24,7 @@ learn how to configure your CA to:
 - [Environment Variables](#environment-variables)
 - [Basic Configuration Options](#basic-configuration-options)
 - [Provisioners](#provisioners)
-- [Certificate Issuance Policy](#policy)
+- [Certificate Issuance Policies](#policy)
 - [Certificate Templates](#templates)
 - [Databases](#databases)
   - [Badger](#badger)
@@ -288,8 +288,17 @@ different provisioners, their target use cases, and how to add, remove, and conf
 
 ## Policy
 
-With certificate issuance policies administrators can configure what Subjects, SANs and Principals the CA is allowed to sign.
-An example of a policy is to only allow (strict) subdomains of `internal.example.com`, which would be encoded as `*.internal.example.com`.
+Certificate issuance policies can be used to enforce which Subjects, SANs and Principals the CA is allowed to sign.
+They can be configured for X.509 certificates as well as SSH user and host certificates.
+Policies are evaluated before a certificate is signed.
+Some examples of policies you can configure are:
+
+- A `dns` rule for `www.example.com`, only matching the domain `www.example.com`
+- A `dns` rule for `*.internal.example.com`, matching all subdomains of `internal.example.com`.
+- An `ip` rule for the `192.168.0.0/24` CIDR, matching all IPs in the range from `192.168.0.0`-`192.168.0.255`.
+- An `email` rule for `@devops`, matching all SSH user principals in the `@devops` domain.
+- A `uri` rule for `*.example.com`, matching all URIs for subdomains of `example.com`, ignoring the URI scheme.
+
 Visit the [`step-ca` policy](/docs/step-ca/policies) page to learn how certificate issuance policies work and how they can be configured.
 
 <Alert severity="info">
@@ -305,19 +314,16 @@ Visit the [`step-ca` policy](/docs/step-ca/policies) page to learn how certifica
   </div>
 </Alert>
 
+
 ## Templates
 
-People use private CAs for all sorts of things, in many different contexts:
-web apps, mobile apps, code signing, cloud VM instances, SSH, IoT devices, etc.
-So `step-ca` must be flexible enough to handle a wide variety of flows.
+Using X.509 and SSH certificate templates administrators can configure information that gets processed and added to certificates.
+These include things like:
 
-X.509 and SSH certificate templates open up these possibilities.
-With certificate templates, you can do things like:
-
-- Add custom SANs or extensions to X.509 certificates
-- Make longer certificate chains, with multiple intermediate CAs
-- Use SSH `force-command` or `source-address` extensions
-- Add conditionals around a certificate's parameters, and fail if they are not met
+- Custom SANs or extensions for X.509 certificates
+- Create longer certificate chains, including multiple intermediate CAs
+- Embed SSH `force-command` or `source-address` extensions
+- Evaluate basic logic operations on certificate's parameters, and fail if requirements are not met
 
 Visit the [`step-ca` templates](/docs/step-ca/templates) page to learn how to use templates.
 
--- a/src/pages/docs/step-ca/configuration.mdx
+++ b/src/pages/docs/step-ca/configuration.mdx
@@ -98,6 +98,19 @@ CA's keys.
       "minUserSSHCertDuration": "5m",
       "maxUserSSHCertDuration": "24h",
       "defaultUserSSHCertDuration": "16h",
+      "policy": {
+        "x509": {
+          "allow": ["*.local"]
+        },
+        "ssh": {
+          "user": {
+            "allow": ["@local"]
+          },
+          "host": {
+            "allow": ["*.local"]
+          }
+        }
+      }
     },
   	"provisioners": [
   		{
--- a/src/pages/docs/step-ca/configuration.mdx
+++ b/src/pages/docs/step-ca/configuration.mdx
@@ -24,6 +24,7 @@ learn how to configure your CA to:
 - [Environment Variables](#environment-variables)
 - [Basic Configuration Options](#basic-configuration-options)
 - [Provisioners](#provisioners)
+- [Certificate Issuance Policy](#policy)
 - [Certificate Templates](#templates)
 - [Databases](#databases)
   - [Badger](#badger)
@@ -284,6 +285,25 @@ CA's keys.
 Provisioners are people or entities that are registered with the certificate authority and 
 authorized to issue certificates. Visit the [`step-ca` provisioners](/docs/step-ca/provisioners) page to learn about the 
 different provisioners, their target use cases, and how to add, remove, and configure them.
+
+## Policy
+
+With certificate issuance policies administrators can configure what Subjects, SANs and Principals the CA is allowed to sign.
+An example of a policy is to only allow (strict) subdomains of `internal.example.com`, which would be encoded as `*.internal.example.com`.
+Visit the [`step-ca` policy](/docs/step-ca/policies) page to learn how certificate issuance policies work and how they can be configured.
+
+<Alert severity="info">
+  <AlertTitle>Need more control?</AlertTitle>
+  <div>
+    A self-hosted <Code>step-ca</Code> instance can be configured with a policy on the authority level only.
+    In a <Link
+      external
+      href="https://smallstep.com/signup?product=cm"
+    >
+      free hosted smallstep Certificate Manager authority
+    </Link>, policies can be configured on the authority, on each provisioner, and on every ACME account.
+  </div>
+</Alert>
 
 ## Templates
 
--- a/src/pages/docs/step-ca/policies.mdx
+++ b/src/pages/docs/step-ca/policies.mdx
@@ -23,7 +23,7 @@ This means that rules for X.509 certificates will not be evaluated when an SSH u
       href="https://smallstep.com/signup?product=cm"
     >
       free hosted smallstep Certificate Manager authority
-    </Link>, policies can be configured on the authority, on each provisioner, and on every ACME account.
+    </Link>, policies can be configured on the authority, on each provisioner, and on every ACME account, providing more granular control.
   </div>
 </Alert>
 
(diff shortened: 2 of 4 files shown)


hslatman:
@tashian: could you have a look at this to see if the docs make sense for someone new to the functionality? I tried to be concise, but to also not miss important aspects.

I've thought a bit about where to put additional info about the additional policy levels for customers. A separate section in the SSH and CM menus didn't seem to really suit, but maybe it works for the CM one.

src/pages/docs/step-ca/configuration.mdx
```diff
@@ -24,6 +24,7 @@ learn how to configure your CA to:
 - [Environment Variables](#environment-variables)
 - [Basic Configuration Options](#basic-configuration-options)
 - [Provisioners](#provisioners)
+- [Certificate Issuance Policy](#policy)
```
tashian: This got me wondering if both the Policies and Templates sections on this page should be promoted to the Core Concepts page, since these two sections are more like intros to the feature, rather than configuration guides.

src/pages/docs/step-ca/policies.mdx
```diff
@@ -0,0 +1,295 @@
+---
+title: Certificate Issuance Policies
+html_title: Certificate Issuance Policies
+description: Certificate issuance policies can be used to allow or deny certificates to be signed based on the requested names.
+---
+
+Administrators can create a certificate issuance policy to configure which certificate names the CA is allowed to sign.
```
tashian, line 7: This feature is related but very different from Name Constraints, and I wonder if we should mention Name Constraints (currently only covered as a template example on https://smallstep.com/docs/step-ca/templates#advanced-x509-template-examples), how policies differ from Name Constraints, and how the user might choose between them—or use them together!

src/pages/docs/step-ca/configuration.mdx
```diff
@@ -285,6 +286,25 @@ Provisioners are people or entities that are registered with the certificate aut
 authorized to issue certificates. Visit the [`step-ca` provisioners](/docs/step-ca/provisioners) page to learn about the 
 different provisioners, their target use cases, and how to add, remove, and configure them.
 
+## Policy
+
+With certificate issuance policies administrators can configure what Subjects, SANs and Principals the CA is allowed to sign.
+An example of a policy is to only allow (strict) subdomains of `internal.example.com`, which would be encoded as `*.internal.example.com`.
```
tashian: Hmm we may want to update the blog post about templates, specifically [this example](https://smallstep.com/blog/x509-certificate-flexibility/#bonus-filter-requests-by-domain-name) of how to do "policies" before we had policies.

src/pages/docs/step-ca/configuration.mdx
```diff
@@ -285,6 +286,25 @@ Provisioners are people or entities that are registered with the certificate aut
 authorized to issue certificates. Visit the [`step-ca` provisioners](/docs/step-ca/provisioners) page to learn about the 
 different provisioners, their target use cases, and how to add, remove, and configure them.
 
+## Policy
```
tashian, line 302: nice intro :D

src/pages/docs/step-ca/policies.mdx
```diff
@@ -0,0 +1,295 @@
...
+- Subject Common Names
+
+The following subsections describe the specifics for each of those.
+
+### Domain Names
+
+Rules for DNS domains are evaluated to match largely according to how [Name Constraints (RFC 5280)](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.10)
+are evaluated, but applied to leaf certificates instead of CA certificates:
+
+- Domain names are normalized into their [internationalized representation (RFC 5890)](https://datatracker.ietf.org/doc/html/rfc5890).
```
tashian: would it make sense to give an example of an internationalized domain name here?

src/pages/docs/step-ca/policies.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+
+
+### URIs
+
+Rules for evaluating URIs are evaluated as follows:
+
+- URI domains are normalized into their [internationalized representation (RFC 5890)](https://datatracker.ietf.org/doc/html/rfc5890).
+- URI rules must contain an authority part and cannot contain IP addresses, in correspondence with [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.10).
+- URI schemes are not matched. When configuring an URI rule, the scheme is ignored.
```
tashian: Hmm. Could this be considered subtly insecure in some way?

src/pages/docs/step-ca/policies.mdx
```diff
@@ -0,0 +1,295 @@
...
+--------|:-:                |:-:                                    |:-:
+cn      |Custom CA Name     |Custom CA Name                         |Different CA Name
+
+
+## SSH Policies
+
+Policies for SSH certificates consist of rules for user and host certificates.
+A policy for SSH user certificates can have rules for the following:
+
+- Email address domains
```
tashian, line 193: Nice. this used to be called `allowedDomains` on the provisioner? Are we keeping that config parameter around?

src/pages/docs/step-ca/policies.mdx
```diff
@@ -0,0 +1,295 @@
...
+- Principals
+
+A policy for SSH host certificates can have rules for the following:
+
+- DNS domains
+- IP ranges and addresses
+- Principals
```
tashian: What's an example of a host cert principal that's not a DNS domain or IP address/range?

src/pages/docs/step-ca/policies.mdx
tashian: I see now that provisioner-level policies are not allowed in open source.

src/pages/docs/step-ca/policies.mdx
```diff
@@ -0,0 +1,295 @@
...
+email       |@devops            |jane@devops                            |john@finance  
+principal   |*                  |johndoe, john, janedoe, jane           |
+principal   |johndoe            |johndoe                                |janedoe                
+
+
+## Policy Configuration
+
+Policies can be administered using the `step` CLI application.
+The commands are part of the [`step ca policy`](https://smallstep.com/docs/step-cli/reference/ca/policy) namespace.
+There are subcommands for administering `authority`, `provisioner` and `acme` policies.
```
tashian: - Since this doc is for open source, I think we should only mention authority policies outside of the CM call-to-action alert box below. 

- Then, in the CLI help text, I think we should make it clear that `provisioner` and `acme` policies are only available in Certificate Manager. We will get questions from ppl about "how do I turn on provisioner policies" if it's not clear from the help that you can't.

- (Same deal with ACME EAB subcommands?)

src/pages/docs/step-ca/policies.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+<CodeBlock language="shell-session" copyText="step ca policy authority ssh host allow dns '*.local'">
+  {`$ step ca policy authority ssh host allow dns "*.local"`}
+</CodeBlock>
+
+You can find more examples in the `step` CLI command reference for the [`step ca policy`](https://smallstep.com/docs/step-cli/reference/ca/policy) namespace.
+
+### Policy in Configuration File
+
+When running a self-hosted `step-ca` it is also possible to configure the authority policy in the `ca.json` CA configuration file.
```
tashian, line 274: may want to mention sending a `HUP` signal after changing the config

tashian (changes requested):
@hslatman looking good! added some comments.

src/pages/docs/step-ca/policies.mdx
hslatman: Great point. I didn't want to mention it (too much) to prevent potential confusion, but I think it does make sense to reference it in some way to show that we based it on some known semantics. People could look it up for some more background info.

Would making the fact that it isn't the same as Name Constraints more explicit make it more clear?

src/pages/docs/step-ca/policies.mdx
hslatman: Yes, I think people may be slightly surprised by this and it could result in a security issue. I remember we discussed a more complete form of URI policy evaluation that would include URNs (a.o.), but decided not to implement it at first. This is probably a good candidate for the skull and bones? 🙂 

It may not be that much work to check the scheme if it was present in in the rule, but I'll need to check this. The harder thing is determining the broader semantics of this: is a scheme optional in the rule? If not, do we need a wildcard for the scheme? Do we also want to support other parts of the URI, like the path, query and fragment?

src/pages/docs/step-ca/policies.mdx
hslatman: I think you're referring to the parameter on the OIDC provisioner? There's an internal note about that, but there it's called `domains` instead. They're different things, restricting the credential itself vs. what names are requested, so that configuration is kept.

The parameter is on a different nesting level and managed differently, so I don't know if it's going to confuse things.

src/pages/docs/step-ca/policies.mdx
hslatman: Well spotted! I concluded the same trying to come up with an example, but forgot to remove it in this list 😅

src/pages/docs/step-ca/policies.mdx
tashian: Maybe! We can try it. I guess I was imagining a couple sentences—maybe a callout—that help people compare Name Constraints to CA policies. If I am designing a PKI, how do I choose where to write and enforce my policies?

src/pages/docs/step-ca/configuration.mdx
hslatman: That absolutely makes sense.

Want a slightly different text here on the configuration page?

src/pages/docs/step-ca/policies.mdx
tashian: ☠️ ☠️ ☠️

src/pages/docs/step-ca/configuration.mdx
tashian: Good question!

I think of this config page as a kind of map. At the top it shows an example config file with ALL possible params, and then it describes those params (and variants on them, like various database drivers or key management services), and it sometimes links off to other pages for more detail if a particular section needs it.

src/pages/docs/step-ca/policies.mdx
hslatman: Added in 2a68d5e6426ee6be71765e750ccd805d075ae293.

src/pages/docs/step-ca/policies.mdx
hslatman: Removed the `provisioner` and `acme` line in 2a68d5e6426ee6be71765e750ccd805d075ae293. 

Working on a PR that includes some documentation improvements here: https://github.com/smallstep/cli/pull/682. This adds a note on the `provisioner` and `acme` subcommands indicating availability in CM. Not on all the subcommands (e.g. for `dns`, `ip`, etc) below that (yet), though. That would require a bit more work due to how these commands are setup in terms of sharing logic.

For ACME EAB we return an error from the CA that indicates availability in CM only when a request is performed against one of the EAB endpoints. The exact same approach is a bit harder to implement for policy, but I think it makes sense to support something similar. I'll have a look at this.

src/pages/docs/step-ca/policies.mdx
hslatman: Removed `Principals` in 2a68d5e6426ee6be71765e750ccd805d075ae293

src/pages/docs/step-ca/policies.mdx
hslatman: Move this bullet to a separate paragraph and prefixed it with ☠️ in 2a68d5e6426ee6be71765e750ccd805d075ae293

src/pages/docs/step-ca/policies.mdx
hslatman: Added a short paragraph about this in 2a68d5e6426ee6be71765e750ccd805d075ae293.

src/pages/docs/step-ca/policies.mdx
hslatman: Added an info block to describe the main difference in 7e51969735952d18ee56fdc475d6142796a9835c. 

Let me know if this info is sufficient. If there's a need to expand on this, I think it should be moved out of the block and to a different page/section. 

Besides our "certificate issuance policies", there's also the core X.509 concept of policies that people could confuse issuance policies with. I think those are not used that much, but I may be in the wrong bubble 😅

src/pages/docs/step-ca/configuration.mdx
hslatman: In b5bbb7d9790c18dde60364842c47cb3de98155e5 I've moved some content to the `Core Concepts` section. Kept the number of examples limited there for policies. In the `Configuration` section I added some more policy examples. Both still forward a reader to more in-depth pages. 

I could add a policy example to the `ca.json` at the top, but that would get quite long if we want to show all keys, so I omitted that for now. Maybe a shortened version like this could work instead:

```json
      "policy": {
        "x509": {
          "allow": ["*.local"]
        },
        "ssh": {
          "user": {
            "allow": ["@local"]
          },
          "host": {
            "allow": ["*.local"]
          }
        }
      }
```

tashian (approved):
lgtm

PR #118, opened 2022-07-20
Wire up popover and tweak styles

alanchrt:
search-tweaks

### DIFF
--- a/src/components/Search/index.jsx
+++ b/src/components/Search/index.jsx
@@ -1,13 +1,7 @@
 import algoliasearch from 'algoliasearch/lite';
 import { default as React, useState, useMemo } from 'react';
 import { InstantSearch } from 'react-instantsearch-dom';
-import {
-  ThemeProvider,
-  makeStyles,
-  Box,
-  List,
-  ClickAwayListener,
-} from '@material-ui/core';
+import { makeStyles, Box, List, Popover } from '@material-ui/core';
 import SearchBox from './search-box';
 import SearchResult from './search-result';
 
@@ -16,19 +10,8 @@ const useStyles = makeStyles({
     background: 'white',
   },
   popover: {
-    background: 'white',
-    position: 'absolute',
-    zIndex: 2,
-  },
-  resultsBorder: {
-    borderRadius: 3,
-    borderColor: '#D3D3D3',
-  },
-  search: {
-    maxHeight: 750,
-    width: 'auto',
-    height: 'auto',
-    boxShadow: 5,
+    maxWidth: 420,
+    maxHeight: 580,
     overflow: 'auto',
     overflowX: 'hidden',
   },
@@ -36,6 +19,7 @@ const useStyles = makeStyles({
 
 export default function Search({ indices }) {
   const [query, setQuery] = useState();
+  const [anchorEl, setAnchorEl] = useState(null);
   const searchClient = useMemo(
     () =>
       algoliasearch(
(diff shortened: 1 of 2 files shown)


src/components/Search/index.jsx
```diff
@@ -44,44 +28,45 @@ export default function Search({ indices }) {
...
+          anchorEl={anchorEl}
+          onClose={handleClose}
+          anchorOrigin={{
+            vertical: 'bottom',
+            horizontal: 'left',
+          }}
+          disableAutoFocus
+          disableEnforceFocus
+          classes={{ paper: classes.popover }}
+        >
```
alanchrt, line 59: Here's the material-ui `Popover` implementation! `anhorEl` refers to the search box, so it can position itself left bottom of the search box.

src/components/Search/search-result.jsx
```diff
@@ -1,104 +1,94 @@
...
   Highlight,
-  Hits,
   Index,
   Snippet,
 } from 'react-instantsearch-dom';
-import { Box, ListItem, ListItemText } from '@material-ui/core';
+import { Box, Typography, ListItem, ListItemText } from '@material-ui/core';
 
-const useStyles = makeStyles({
+const useStyles = makeStyles((theme) => ({
```
alanchrt, line 14: `makeStyles` accepts a function with `theme` as an argument, so we can use variables from our step-ui theme.

src/components/Search/search-result.jsx
```diff
@@ -1,104 +1,94 @@
...
   resultsBorder: {
-    borderColor: '#D3D3D3',
-    marginBottom: -16,
-    marginTop: -2,
+    borderBottom: `1px solid ${theme.palette.accent.grey3}`,
```
alanchrt, line 16: This is a more standard color in our theme instead of defining new ones.

src/components/Search/search-result.jsx
```diff
@@ -1,104 +1,94 @@
...
     '& mark ': {
-      backgroundColor: '#84A8FF',
-    },
-    '& ul ': {
-      listStyle: 'none',
-    },
-    '& a:-webkit-any-link ': {
-      textDecoration: 'none',
-      color: 'inherit',
+      backgroundColor: theme.palette.accent.yellow,
```
alanchrt, line 24: Same, a standard color instead of new one.

src/components/Search/search-result.jsx
```diff
@@ -1,104 +1,94 @@
...
 
 const HitCount = connectStateResults(({ searchResults }) => {
   const hitCount = searchResults && searchResults.nbHits;
-  const theme = useStyles();
+  const classes = useStyles();
   return hitCount > 0 ? (
-    <Box borderBottom={1} textAlign={'right'} className={theme.resultsBorder}>
-      <Box mr={2}>
+    <Box textAlign={'right'} className={classes.resultsBorder} mb={2}>
+      <Typography variant="body1">
```
alanchrt, line 33: I noticed that react was throwing warnings because `Paragraph` uses a `p` tag and it nests weirdly with some of the algolia stuff (and adds extra margin too). It ends up working best to just use the `Typography` component directly.

src/components/Search/search-result.jsx
```diff
@@ -1,104 +1,94 @@
...
           }
         />
       </ListItem>
     </Link>
-  </Box>
+  );
+};
+
+const Hits = connectHits(({ hits }) =>
+  hits.map((hit) => <PageHit key={hit.slug} hit={hit} />)
```
alanchrt, line 75: I found this thing from algolia where you can make your own custom `Hits` component instead of trying to target theirs with styles (weird negative margins and stuff).

src/components/Search/search-result.jsx
```diff
@@ -1,104 +1,94 @@
...
           : `/docs/step-cli/reference/${hit.slug}`
       }
+      className={classes.link}
     >
-      <ListItem button attribute="slug" hit={hit} tagName="mark">
+      <ListItem button attribute="slug" hit={hit} className={classes.hits}>
         <ListItemText
           primary={
-            <h4>
+            <Heading variant="h6">
```
alanchrt, line 59: `h4` adds some default margins that are awkward, so we have a `Heading` component to standardize the styling a bit.

src/components/Search/search-result.jsx
```diff
@@ -24,7 +24,6 @@ const useStyles = makeStyles((theme) => ({
       backgroundColor: theme.palette.accent.yellow,
     },
   },
-}));
```
alanchrt, line 42: Looks like this line accidentally got lost in the merge.

PR #124, opened 2022-08-08
X5C Provisioner: add note regarding validity boundaries

Slamdunk:
See https://github.com/smallstep/certificates/issues/985


tashian:
Hi @Slamdunk after reading the related issue, I assume this is no longer needed. If you feel something is missing here, please reopen the issue and let me know. Thanks!

Slamdunk:
Hi @tashian, I think it's not immediately clear that Certs validity cannot extend over token's one when working with `X5C` provider, so to me this PR is still a valid addition to docs.

tashian:
Thanks @Slamdunk , I've addressed this in eea47d1

PR #126, opened 2022-08-09
Docsearch mobile bug

masoncooper19:
@alanchrt made some changes based on the feedback in linear. the clickaway was removed and replaced with this clear button (the x in the search box) which closes the search results and clears the search bar when clicked. what do you think of these changes?
Screenshot 2022-08-09 135520

### DIFF
--- a/src/components/Search/index.jsx
+++ b/src/components/Search/index.jsx
@@ -8,7 +8,7 @@ import SearchResult from './search-result';
 const useStyles = makeStyles((classes) => ({
   resultsOpen: {
     background: 'white',
-    position: 'absolute',
+    position: 'sticky',
   },
 }));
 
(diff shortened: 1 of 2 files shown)


src/components/Search/index.jsx
```diff
@@ -22,41 +26,54 @@ export default function Search({ indices, changeOpenSearch }) {
       ),
     []
   );
-  const classes = useStyles();
   const [openResults, setOpenResults] = React.useState(false);
+  const [value, setValue] = useState();
   const handleClick = () => {
     changeOpenSearch(true);
     setOpenResults(true);
+    setValue();
```
alanchrt, line 34: Is this necessary?

alanchrt (approved):
Yes, this is awesome!

src/components/Search/index.jsx
masoncooper19: yes because without this function then value="" permanently, thus making it so you cannot type in any more searches

PR #127, opened 2022-08-15
Add section to OIDC provisioner about Device Authz Grant flow

dopey:


src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -609,6 +609,38 @@ Bottom line, the OAuth flow implemented in `step` and `step-ca` is widely vetted
   </div>
 </Alert>
 
+#### Device Authorization Grant Flow
```
tashian: I think we should move this section above "Notes" and call it something like "Browserless Console Mode"

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -609,6 +609,38 @@ Bottom line, the OAuth flow implemented in `step` and `step-ca` is widely vetted
...
+
+The Device Authorization Grant flow is an OAuth 2.0 extension that enables OAuth for devices with no browser 
+or limited input capability to obtain an access token.
+The OIDC provisioner supports the Device Authorization Grant flow, however,
+it is not always the default flow for a given OIDC provider.
```
tashian:
```suggestion
Sometimes it's helpful to use OAuth in an input-constrained environment where a web browser is not available.
The Device Authorization Grant flow is an OAuth 2.0 extension designed for this scenario.
The OIDC provisioner supports the Device Authorization Grant flow.
However, it is not always the default flow for a given OIDC provider.
```

When you say it's not "always" the default—is it the default flow in some cases? (That is, can I configure the provisioner to always have its clients use the Device Authorization Grant flow?)

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -609,6 +609,38 @@ Bottom line, the OAuth flow implemented in `step` and `step-ca` is widely vetted
...
+
+The Google OIDC provider initially defaulted to a different flow - the Out Of Band (OOB) flow which has since been deprecated. 
+`step-ca` OIDC provisioners configured to use Google OIDC will still default to the OOB flow (until October 3rd, 2022).
+Users can specify the proper OAuth 2.0 flow by supplying flags and environment variables.
```
tashian: Can we make this a little warning callout rather than part of the main text?

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -609,6 +609,38 @@ Bottom line, the OAuth flow implemented in `step` and `step-ca` is widely vetted
...
+
+To use the the default OAuth 2.0 flow for input constrained devices (for your OIDC provider) use:
```
tashian:
```suggestion
To use the Device Authorization Grant OAuth 2.0 flow for input-constrained devices with an OIDC provisioner, run:
```

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -609,6 +609,38 @@ Bottom line, the OAuth flow implemented in `step` and `step-ca` is widely vetted
...
+
+```shell-session
+$ step ca certificate foo foo.crt foo.key --console
+```
+
+For `step oauth`, specify a non-default OAuth 2.0 input constrained flow with the `--console-flow device` flag and value:
```
tashian:
```suggestion
For `step oauth`, request an OAuth 2.0 input-constrained flow with the `--console-flow device` flag and value:
```

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -609,6 +609,38 @@ Bottom line, the OAuth flow implemented in `step` and `step-ca` is widely vetted
...
+
+```shell-session
+$ step oauth --client-id <my-client-id> --client-secret <my-client-secret> --console-flow device
+```
```
tashian: This is the first time we're mentioning `step oauth` on this page, and it may be out of scope to put this here, since this is the provisioner page.

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -609,6 +609,38 @@ Bottom line, the OAuth flow implemented in `step` and `step-ca` is widely vetted
...
+
+For any other `step` command, specify a non-default OAuth 2.0 input constrained flow with the `STEP_CONSOLE_FLOW=device` environment variable ahead of the command:
+
+```shell-session
+$ STEP_CONSOLE_FLOW=device step ca certificate foo foo.crt foo.key
+```
+
```
tashian: This may also be out of scope for this page, since we cover `STEP_` variables in the Step CLI docs. But, is there a special reason people might want to know they can use the env variable option here?

src/pages/docs/step-ca/provisioners.mdx
```diff
@@ -609,6 +609,38 @@ Bottom line, the OAuth flow implemented in `step` and `step-ca` is widely vetted
...
+After October 3rd, 2022 Google will deprecate the OOB flow for existing clients and 
+the Device Authorization Grant flow will become the default flow for all input constrained devices.
```
tashian: This can be part of the above callout too I think

tashian (changes requested):
hey just a few notes on this one. lookin good

src/pages/docs/step-ca/provisioners.mdx
dopey: It is the default flow for everything except Google at this time. If you use the Google Identity for your provisioner you would need to add the `STEP_CONSOLE_FLOW=device` env var to your defaults.json to get it to use the device flow by default.

src/pages/docs/step-ca/provisioners.mdx
dopey: The problem is that this isn't currently true. It will always be true after October 3rd. It's currently only true for non-Google providers. We can just say whatever, it will be true soon.

src/pages/docs/step-ca/provisioners.mdx
dopey: removed the oauth example.

src/pages/docs/step-ca/provisioners.mdx
dopey: `step ca xxx` do not accept the console_flow flag. Only `step oauth` accepts that flag. So you need to use the env var for `step ca xxx` if you want to change the default flow.

src/pages/docs/step-ca/provisioners.mdx
dopey: removed this section.

Issue #128, opened 2022-08-22
Azure Key Vault Authentication should prefer Managed Identity vs usage of Service Principals.


rjfmachado:
Using a Managed Identity for a Virtual Machine may be preferable to Service Principals as the secret does not have to be directly managed and exposed to the environment (https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication-managed-identity?tabs=azure-cli)

I've tested step ca and cli using Managed Identity to provision a standalone CA and run step-ca as a daemon with key vault with no issues and configuration can be reduced to:
* configure a managed identity (either system/user assigned) and assign the necessary RBAC/access policy rights to Key Vault.
* in the guest OS, configure the AZURE_CLIENT_ID to the client id of the Managed Identity. AZURE_TENANT_ID is optional.

see https://github.com/rjfmachado/step-ca-azure, values are replaced by the template deployment.
* using AZURE_CLIENT_ID=[AZURE_CLIENT_ID] in /etc/environment for the cli part
* using Environment=AZURE_CLIENT_ID=[AZURE_CLIENT_ID] in /etc/systemd/system/step-ca.service for the daemon.

The Azure SDK for go will use the managed identity if the secret is not presented.
https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/default_azure_credential.go#L31

tashian:
Thanks for this. I've merged the PR and I also added your `step-ca-azure` example to our [list of community-contributed examples](https://github.com/smallstep/certificates/discussions/765).

PR #129, opened 2022-08-22
Add Managed Identity guidance

rjfmachado:
As is, step cli and step-ca already support Azure Managed Identities and they should be preferable to Service Principals.
I've created #128 with more details about the changes.


tashian:
Thanks @rjfmachado for the clarity!

PR #141, opened 2022-09-08
Document webhooks

areed:


tashian:
Hi @areed, what's the status on webhooks in step-ca? Is this PR ready for review/merge?

areed:
> Hi @areed, what's the status on webhooks in step-ca? Is this PR ready for review/merge?

I've updated and requested a review. The feature will be released in step-ca 0.23.

src/pages/docs/step-ca/webhooks.mdx
```diff
@@ -0,0 +1,80 @@
+---
+title: Webhooks
+html_title: Webhooks
+description: Webhooks augment the data available to certificate templates
+---
+
+Administrators can attach webhooks to provisioners to retrieve additional data that will be available when rendering [certificate templates](/docs/step-ca/templates).
```
tashian: Could we add a callout at the top here, making it clear that the feature requires enabling Remote Provisioner Management, with a link to `/docs/step-ca/provisioners/#remote-provisioner-management`

src/pages/docs/step-ca/webhooks.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+```bash
+step ca provisioner webhook add my_provisioner my_webhook --url https://example.com
```
tashian: How about `https://localhost:9443` here since the what the example server binds to?

src/pages/docs/step-ca/webhooks.mdx
```diff
@@ -0,0 +1,80 @@
...
+```
+
+A template on the provisioner will then be able to reference the response under the path `.Webhooks.webhook_name`.
+If the webhook was named `people`, the role in the webhook response could be accessed in a template under the field `.Webhooks.people.role`.
+
+## Requests
+
+All requests will use the POST method to send a json body to the webhook server containing a `timestamp` field.
+Additional data will vary based on the type of the certificate being signed.
+Refer to the [webhooks server example repository](https://github.com/smallstep/webhooks) for examples in Go of parsing webhook request bodies for both x509 and SSH certificate requests.
```
tashian: The example server should:
- have instructions for building and running the server (for non-Go folks)
- remove the `replace`s from go.mod (once 0.23 is out I guess?)
- describe the server certs `webhook.crt` and `webhook.key` and  show them how to run `step ca certificate localhost webhook.crt webhook.key` in the readme

src/pages/docs/step-ca/webhooks.mdx
tashian: I think this command shouldn't be right at the top of the doc here. For order of operations, my webhook server need to already be running before I run this command, otherwise it creates a condition where:
- Running `step ca provisioner update` yields a `Post "https://localhost:5555": dial tcp [::1]:5555: connect: connection refused.`, so I can't edit the provisioner
- I can't start the webhook server without a server cert, but I can't run `step ca certificate` anymore

src/pages/docs/step-ca/webhooks.mdx
tashian: Webhooks could also be used to gather CSR metadata for other purposes, eg. metrics or indexing certificate requests, and we may want to mention that here.

Is it possible for webhooks to be used to monitor issued certs as well?

(Are webhooks the last approval step when a CSR is being signed? For example, if there's logic in the template that might cause the CSR to be rejected by step-ca, will that rejection have already happened by the time the webhook is called?)

tashian: On the other hand if webhooks in open source are really limited in terms of their use—if they really are just an extension of templates—then we should consider placing these docs under the Templates doc, rather than in a separate page.

tashian (changes requested):
Thanks @areed I think this covers all the basics.

Feels like this page would benefit from a tutorial section that allows someone to try out the feature locally, quickly. The tutorial would leave them with a working environment in which to develop their own webhook. I think if we extracted & elaborated on the webhooks server repo's README, that would do the trick.

I tried testing this feature out and I still don't have it working locally.
I definitely fired off a couple of footguns along the way, which I'll share here:

Here's what I tried the first time:

- `step ca init --remote-management`
- start `step-ca`
- run `step ca provisioner webhook add carl@smallstep.com hook --url https://localhost:5555`
- copy `$(step path)/certs/root_ca.crt` to webhook server dir and update the webhook server to read it
- discover that I needed a server cert for the webhooks server
- at this point I couldn't do anything else, because the default JWT provisioner was trying to access the webhook.

My second attempt looked like this:

- `step ca init --remote-management`
- start `step-ca`
- copy `$(step path)/certs/root_ca.crt` to webhook server dir
- run `step ca certificate localhost webhooks.crt webhooks.key`
- update webhook server to use `localhost:5555` and start webhook server
- run `step ca provisioner webhook add carl@smallstep.com hook --url https://localhost:5555`
- update webhook server with the new webhook ID and secret, and restart it
- run `step ca provisioner update carl@smallstep.com --x509-template ./leaf.tpl` to add a template that uses the webhook
- this returns "`webhook server did not allow request.`", and the webhook server prints "`Received x509 enriching webhook request for "". Sent data: {Role:}`"
- Try to backtrack and remove the webhook server by running `step ca provisioner webhook remove carl@smallstep.com hook`, but that returns the same errors.

My third attempt looked like this:

- `step ca init --remote-management`
- start `step-ca`
- copy `$(step path)/certs/root_ca.crt` to webhook server dir
- run `step ca certificate localhost webhooks.crt webhooks.key`
- `step ca provisioner add webhooktest --x509-template ~/projects/20221108\ webhooks/leaf.tpl --create`
- update webhook server to use `localhost:5555` and start webhook server
- run `step ca provisioner webhook add webhooktest hook --url l https://localhost:5555`
- update webhook server with the new webhook ID and secret. add `carl@smallstep.com` with role `eng` to the `db`.
- start webhook server
- run `step ca certificate carl@smallstep.com carl.crt carl.key --provisioner webhooktest`
- `webhook server did not allow request.`
- at least now I can still edit my template and provisioner! :D

src/pages/docs/step-ca/webhooks.mdx
```diff
@@ -0,0 +1,80 @@
...
+step ca provisioner webhook update my_provisioner my_webhook --bearer-token=abc123xyz
+```
+
+or for basic authentication:
+
+```
+step ca provisioner webhook update my_provisioner my_webhook --basic-auth-username user --basic-auth-password pass
+```
+
+#### TLS Client Certificate Authentication
```
tashian: this is a nice feature

src/pages/docs/step-ca/webhooks.mdx
tashian: I see now that webhooks are useful for two purposes:
- enriching certs
- external policy/authorization evaluation

and maybe a third: metrics & auditing hooks.

I think it'd help to describe these use cases briefly in the intro.

areed: Callout added in 53acabd05a61a131bba8be9a002b7188b7cb4a51. There was a product decision made to not document the policy/authorization capabilities of webhooks yet because it could create confusion of how that interacts with other features for enforcing policies. The default webhook kind is ENRICHING, which happens early and so includes the CSR. The other kind is AUTHORIZING and that is called at the end with the certificate to be signed.

areed:
I added a tutorial in 53acabd05a61a131bba8be9a002b7188b7cb4a51

src/pages/docs/step-ca/webhooks.mdx
areed: Added [requirements](https://github.com/smallstep/webhooks#requirements) and removed replace directives.

src/pages/docs/step-ca/webhooks.mdx
tashian: I see.

tashian (commented):
Hi @areed, I made some updates for clarity here.

Some small questions for you:

- Should we address smallstep/cli#803 here, or in the CLI itself?

- Do you expect the flag for disabling mutual TLS to be commonly used? Given that it's up to the webhook server to decide whether to enforce mutual TLS, we may not need to document this flag explicitly.

- I'm wondering the same thing about the `Authorization` header. If the Signature + Mutual TLS is the golden path we expect people to use here, maybe we can reduce the section on the Authorization header to a mention.

- For the signature, is it necessary for the webhook server to confirm the signature if mutual TLS is being used, or, in that case, would it be safe for the webhook server to just use `X-Smallstep-Webhook-ID` to associate the request, and ignore `X-Smallstep-Signature`? I recognize that the webhook server would have to inspect the client certificate more closely in order for this to be a feasible alternative to the signature.

areed:
> Should we address https://github.com/smallstep/cli/issues/803 here, or in the CLI itself?

We can discuss at open source triage tomorrow.

> Do you expect the flag for disabling mutual TLS to be commonly used? Given that it's up to the webhook server to decide whether to enforce mutual TLS, we may not need to document this flag explicitly.

No, I don't expect that to be commonly used. The only use case is if you don't want the webhook server to know the authority's domain name.

> I'm wondering the same thing about the Authorization header. If the Signature + Mutual TLS is the golden path we expect people to use here, maybe we can reduce the section on the Authorization header to a mention.

We expected the Signature to be the golden path for custom webhook servers and added Authorization headers and client certificates for broader compatibility.

> For the signature, is it necessary for the webhook server to confirm the signature if mutual TLS is being used, or, in that case, would it be safe for the webhook server to just use X-Smallstep-Webhook-ID to associate the request, and ignore X-Smallstep-Signature? I recognize that the webhook server would have to inspect the client certificate more closely in order for this to be a feasible alternative to the signature.

Enriching webhooks are called before the provisioner is added to the certificate to be signed so there's no way to verify which provisioner is calling the webhook if you ignore the Signature header. There may be cases where you have metadata that can be shared with some provisioners but not others.

src/pages/docs/step-ca/webhooks.mdx
```diff
@@ -0,0 +1,219 @@
...
+### Authentication
+
+Your webhook server must authenticate each request from the CA.
+To achieve this, the CA sends a signature of its payload in the webhook request header,
+and the webhook server must confirm this signature.
+The signature also tells the webhook server which webhook is currently being executed.
+
+In addition the the signature header, you can optionally enable two other authentication schemes:
+- **Authentication Header**  
+  The CA can send a bearer token or username and password in an `Authentication` header.
```
areed, line 166: Should this be `Authorization` header?

tashian: Good catch 🙌

tashian (approved):

PR #143, opened 2022-09-13
minor wording modification for step-ca sign command

Linda-Ikechukwu:


tashian:
@Linda-Ikechukwu unfortunately, we can't edit these reference pages directly, because they are automatically generated by the `step` CLI command. Everything under `src/pages/docs/step-cli/reference` is auto-generated whenever we release a new version of `step`. So, you'll have to make the edit in [smallstep/cli](https://github.com/smallstep/cli) and make a PR over there. The reference docs are source files under the `command` folder in that repo. I'll update the README on this repo to clarify this caveat! :D

PR #170, opened 2022-11-23
Fix the short policy example

hslatman:
This was reported in https://github.com/smallstep/certificates/issues/1181. The compact example was missing `dns` and `email` labels.


tashian (approved):
lgtm

PR #173, opened 2022-12-07
Add certificate duration to traefik config

dmorawetz:
As stated in the section about [renewal period](https://smallstep.com/docs/tutorials/acme-protocol-acme-clients#renewal-period), one must set the correct valid time so traefik can correctly renew certificates automatically. Adding the correct setting may save a lot of debug time.

#### Name of feature:
Traefik ACME config

by default, Traefik assumes certificate validity for 90 days and does not correctly renew certificates from step-ca with default config.

#### Is there documentation on how to use this feature? If so, where?
https://doc.traefik.io/traefik/https/acme/#certificatesduration

#### In what environments or workflows is this feature supported?
Deployment with Traefik reverse proxy.


tashian:
Good catch, thank you @dmorawetz

tashian (approved):

PR #177, opened 2022-12-21
Update RA  doc to correct reference to CA password

shrikeh:
#### Name of feature:

Documentation update

The previous version incorrectly stated that `--issuer-password-file` was against the CA. It is against the RA. Additionally, a small `docker-compose.yml` extract is provided as an example of how to achieve this using the `step-ca` docker image

#### Is there documentation on how to use this feature? If so, where?
No, there is no documentation on the documentation. Curses!

#### Supporting links/other PRs/issues:

💔Thank you!

### DIFF
--- a/src/pages/docs/step-ca/registration-authority-ra-mode.mdx
+++ b/src/pages/docs/step-ca/registration-authority-ra-mode.mdx
@@ -143,7 +143,25 @@ you can create a new key for your RA by decrypting the CA's JWK provisioner key 
   }
   ```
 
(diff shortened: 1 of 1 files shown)


tashian:
Thanks for the PR on this @shrikeh! I'll get this merged after the holidays when we're all back in the office. I may extract your note about the entrypoint script into our Docker README, as it applies more broadly to flags passed into `step-ca`.

tashian:
Fixed in #179

PR #217, opened 2023-04-28
Update acme-protocol-acme-clients.mdx

williamtrelawny:
Added section on configuring Caddy to use TLS-ALPN-01 challenges instead of HTTP-01 when caddy is bound to port 80.

#### Supporting links/other PRs/issues: https://discord.com/channels/837031272227930163/841243933936975912/1101584170855190638

💔Thank you!

### DIFF
--- a/tutorials/acme-protocol-acme-clients.mdx
+++ b/tutorials/acme-protocol-acme-clients.mdx
@@ -242,6 +242,20 @@ tls carl@smallstep.com {
 }
 ```
 
+If caddy is already bound to port 80 (as reverse proxies typically are), then you **_must_** use TLS-ALPN-01 challenges instead of the default HTTP-01 because HTTP-01 challenges must be performed on port 80 per the [ACME standard](https://letsencrypt.org/docs/challenge-types/) ([RFC-8555](https://datatracker.ietf.org/doc/html/rfc8555#section-8.3)). Here is an example of this (ref: [caddy docs](https://caddyserver.com/docs/caddyfile/directives/tls#acme)):
+
+```
+foo.internal {
+        tls {
+                issuer acme https://ca.internal/acme/acme/directory {
+                        email carl@smallstep.com
+                        disable_http_challenge
+                        trusted_roots <step path>/certs/root_ca.crt
+                }
+        }
+}
+```
+
 Replace `<step path>` with the output of the [`step path`](../step-cli/reference/path/) command.
 
 Now run `caddy` to start serving HTTPS!


tutorials/acme-protocol-acme-clients.mdx
```diff
@@ -242,6 +242,20 @@ tls carl@smallstep.com {
 }
 ```
 
+If caddy is already bound to port 80 (as reverse proxies typically are), then you **_must_** use TLS-ALPN-01 challenges instead of the default HTTP-01 because HTTP-01 challenges must be performed on port 80 per the [ACME standard](https://letsencrypt.org/docs/challenge-types/) ([RFC-8555](https://datatracker.ietf.org/doc/html/rfc8555#section-8.3)). Here is an example of this (ref: [caddy docs](https://caddyserver.com/docs/caddyfile/directives/tls#acme)):
```
tashian, line 245:
```suggestion
In a reverse proxy configuration
where Caddy is bound to port 80,
you [cannot use the `http-01` challenge](https://caddy.community/t/acme-http01-challenge-on-a-non-80-port/16412/2) as described above.
An alternative is to use a `tls-alpn-01` challenge instead,
as shown in this example `Caddyfile` block:
```

tashian: @mholt can you confirm that if Caddy is bound to port 80 already, the `http-01` challenge type cannot be used and must be disabled?

tashian (changes requested):
Thanks for the PR and the context here @willman42. I've added a suggestion and tagged Matt from Caddy just to confirm.

tutorials/acme-protocol-acme-clients.mdx
mholt: If Caddy is serving port 80, then the `http-01` challenge can still be used as long as the ACME challenge request is proxied to the ACME client.

williamtrelawny: True, but in this case I'm using Caddy **_as_** the ACME client. In this situation caddy complains that port 80 is already in use when attempting the http-01 challenge. See the link to the conversation in discord in my OP comment for supporting evidence.

I'll try to get another demo made tonight too.

mholt: @willman42 When I click the Discord link it just says "No text channels" for me :man_shrugging: 

Can you summarize here what the problem is?

If Caddy is already listening on port 80, it will not open port 80 to solve the http challenge, it has to assume that whatever is listening on port 80 can solve it. If Caddy is listening on port 80, it will use that to solve the http challenge as long as it initiated the challenge. If it did not, it will process the request like any other HTTP request, allowing you to proxy it to another ACME client for example.

williamtrelawny: Here is the step-ca error message I get when attempting to do the `http-01` challenge on port 80 to caddy using the identifier `dns.corp.example.com`:
```
Apr 28 14:47:54 docker.lab step-ca[1217480]: time="2023-04-28T14:47:54-04:00" level=info duration=5.777340067s duration-ns=5777340067 fields.time="2023-04-28T14:47:49-04:00" method=POST name=ca nonce=eGsxWTRib3Y1Rk14VTFBQm9NcEIwTFhia0JtaTVkZFg path=/acme/acme/challenge/NPJyTUGuEm86XqBeSwIyskHO622f5AzC/AiQVwjbZohwC8SPXJwTLQkXjIQkCrs5A protocol=HTTP/2.0 referer= remote-address=172.20.0.2 request-id=ch619ldfl9gk4mbkgt90 response="{\"type\":\"http-01\",\"status\":\"pending\",\"token\":\"ImZkJ7I7x2OVeeIOo9HgD0SqyCS8edAM\",\"url\":\"https://ca.corp.example.com:8443/acme/acme/challenge/NPJyTUGuEm86XqBeSwIyskHO622f5AzC/AiQVwjbZohwC8SPXJwTLQkXjIQkCrs5A\",\"error\":{\"type\":\"urn:ietf:params:acme:error:connection\",\"detail\":\"The server could not connect to validation target\"}}" size=336 status=200 user-agent="Caddy/2.6.2 CertMagic acmez (linux; amd64)" user-id=
```
It just repeats the above message complaining it cannot reach the validation target. `step-ca` is running on the same OS as the caddy container. Here is the relevant caddy log too: `{"level":"info","ts":1683058018.3440974,"logger":"http.acme_client","msg":"trying to solve challenge","identifier":"dns.corp.example.com","challenge_type":"http-01","ca":"https://ca.corp.example.com:8443/acme/acme/directory"}`

nslookup on the base OS proving DNS is working for all hostnames involved:
```
will@docker:/etc/step-ca$ nslookup dns.corp.example.com
Server:         10.0.20.10
Address:        10.0.20.10#53

Name:   dns.corp.example.com
Address: 10.0.20.10

will@docker:/etc/step-ca$ nslookup ca.corp.example.com
Server:         10.0.20.10
Address:        10.0.20.10#53

Name:   ca.corp.example.com
Address: 10.0.20.10
```

docker-compose.yml excerpt show caddy is bound to host port 80:
```
services:
  caddy:
    container_name: caddy
    image: caddy:latest
    networks:
      - caddy-net
    restart: always
    ports:
      - "80:80"
      - "443:443"
    volumes:
     - './caddy/Caddyfile:/etc/caddy/Caddyfile'
     #- '$PWD/site:/srv'
     - './caddy/data:/data'
     - './caddy/config:/config'
     - '/etc/step-ca/certs:/certs'

    extra_hosts:
     - "ca.corp.example.com:10.0.20.10"
 ```
 
 Caddyfile:
 ```
 dns.corp.example.com {
        tls admin@example.com {
                ca https://ca.corp.example.com:8443/acme/acme/directory
                ca_root /certs/root_ca.crt
        }
}
```
`/certs` container path is mounted to `/etc/step-ca/certs` host path like you see above. Caddy is not configured to proxy traffic on port 80, so it's handling all port 80 traffic itself.

Am I missing something? I feel like this is the exact reason why you don't use the `http-01` challenge for Traefik too.

mholt: Thanks for the info; but I'm still not sure I understand fully what's going on...

So as I understand it, there's an ACME CA and the client running on the same machine (same IP). The server, in trying to make the port 80 request to prove the http-01 challenge, is reporting an error: `"The server could not connect to validation target"`

The client, Caddy, initiated the http-01 challenge, and emits only (?) one log, which is, `"trying to solve challenge"` -- and then it just stops? Are there no other logs at all? Seeing only 1 in a complex process like an ACME transaction leaves a lot of guesswork. What shows if you enable debug mode?

If you don't see any logs like `"served key authentication"`, then the ACME CA was in fact not able to connect to Caddy. If you make the http-01 challenge request with `curl -v` (or even a request that just _looks_ like the real one) you should see some Caddy logs about the ACME challenge. I suspect the issue is a networking problem (bad routing perhaps).

tashian: Yes, often the error `"The server could not connect to validation target"` is raised because of a either DNS or networking issue, especially when deployed in container environments. The CA needs to be able to resolve and create an outbound connection to the ACME client in order to validate the challenge, and it will raise this error if it can't.

See https://github.com/smallstep/certificates/discussions?discussions_q=The+server+could+not+connect+to+validation+target

mholt: It sounds like this is a networking problem then -- I don't think there should be any reason to disable http-01 (unless you need a different challenge for policy reasons).

williamtrelawny: If it were a networking issue then the `tls-alpn-01` challenge wouldn't work either right?

mholt: Depends if the same rules that are hindering the connection on port 80 also apply to port 443. If not, then the tls-alpn-01 challenge would indeed succeed.

williamtrelawny:
@mholt @tashian you can close this PR. My issue was that the host OS could not resolve the target `dns.corp.example.com` without the use of its upstream DNS server, which was running as a container on the same host behind the Caddy proxy already. So a mix of DNS and networking issues.

Very sorry to waste everyone's time, I thought I was on to something here...

mholt:
@willman42 No worries, thanks for figuring it out!

PR #230, opened 2023-05-11
Link fixes on SSH index page

tashian:
The `<ContentLink>` component has a different behavior for resolving links, that doesn't match the way we do it everywhere else, but still passes the link checker, so I have removed it in favor of a regular Markdown link.

### DIFF
--- a/ssh/README.mdx
+++ b/ssh/README.mdx
@@ -28,17 +28,13 @@ you haven't already.
 
 Start by configuring SSH access on your laptop:
 
-<ContentLink icon={<FileWithFlippedCornerIcon />} href="client">
-  Client Quickstart Guide &gt;
-</ContentLink>
+**[Client Quickstart Guide](./client.mdx)**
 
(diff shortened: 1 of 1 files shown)


ssh/README.mdx
```diff
@@ -28,17 +28,13 @@ you haven't already.
...
+**[Client Quickstart Guide](./client.mdx)**
 
 ## Step 2: Register Hosts
 
 Register your hosts with Smallstep SSH:
 
-<ContentLink icon={<FileWithFlippedCornerIcon />} href="hosts">
-  Host Quickstart Guide &gt;
-</ContentLink>
+**[Host Quickstart Guide](./client.mdx)**
```
andysmallstep: @tashian - Should this be something like (./host.mdx)** versus client? It appears to be the same as the line 31 change to (./client.mdx)**

andysmallstep (approved):

PR #249, opened 2023-06-22
Add Arch Linux `step` symlink to install instructions

tashian:


hslatman (approved):

hslatman:
The reported link issue seems to be a false positive

PR #260, opened 2023-08-07
Make term "trust store" consistent, and suggestions [..]

toraritte:
#### Name of feature:

Term consistency

Makes it possible to find all instances of the term "trust store".

#### Why is this important to the project (if not answered above):

I find consistency and discoverability of concepts important.

#### Supporting links/other PRs/issues:

None that I know of.

---

From commit message:

> The term "system's default  trust store" is used way more  than  "system  truststore"  or  "system  trust store", but  this commit makes it  consistent across the docs.
> 
> Other amendments:
> 
> + use "root certificate"  instead  of  "certificate" where appropriate to avoid ambiguities
> 
> +  use  "supported  trust  stores"  instead  of  the "system's default trust store" where appropriate

Thank you for taking a look!

### DIFF
--- a/certificate-manager/getting-started.mdx
+++ b/certificate-manager/getting-started.mdx
@@ -87,7 +87,7 @@ Run the following command, substituting the values from your Authority's propert
 
 This command will download the CA Root certificate and configure your local `step` client to interact with the Authority.
 
-If desired, you can also use the `step` CLI to [install](../step-cli/reference/certificate/install) the CA Root certificate to your system's truststore.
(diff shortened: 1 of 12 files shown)


tashian:
Hi! Thanks for the PR. Your changes under `step-cli/reference`, however, are in a section that's auto-generated by our CLI's help text, so those will be overwritten on the next release. I'll go ahead and merge this, though. If you want to update the CLI help text in a way that will persist, you'll need to make a PR on [smallstep/cli](https://github.com/smallstep/cli) instead.

tashian (approved):

toraritte:
(For cross reference: PR https://github.com/smallstep/cli/pull/999)

PR #268, opened 2023-09-11
chore: move kubernetes tls content into correct directory structure

C-Mejlak:
This is so the correct element is highlighted in the docs side bar:

Example URL: https://smallstep.com/docs/certificate-manager/kubernetes-ingress-tls/

**Before**

image

**After**

image


tashian (approved):
lgtm

PR #273, opened 2023-09-28
Update install instructions to use unversioned URLs

tashian:


step-cli/installation.mdx
```diff
@@ -176,9 +176,9 @@ Download the Windows binary from our [latest releases](https://github.com/smalls
 Open PowerShell and run the following:
 
 ```shell
-curl.exe -LO https://dl.smallstep.com/gh-release/cli/docs-cli-install/v0.24.4/step_windows_0.24.4_amd64.zip
-Expand-Archive -LiteralPath .\step_windows_0.24.4_amd64.zip -DestinationPath .
-step_0.24.4\bin\step.exe version
+curl.exe -LO https://dl.smallstep.com/cli/docs-cli-install/latest/step_windows_amd64.zip
+Expand-Archive -LiteralPath .\step_windows_amd64.zip -DestinationPath .
+step_0.25.0\bin\step.exe version
```
hslatman: Is this final version number in the path intentional?

hslatman (commented):
This one path maybe need not be versioned. Otherwise: nice 😄

step-cli/installation.mdx
tashian: It is intentional, but on second though we should probably use this opportunity to remove the version number from the wrapped dirname. I can't see why we wouldn't do that. 
So I created these two PRs to do that:
https://github.com/smallstep/cli/pull/1034
https://github.com/smallstep/certificates/pull/1580

If we agree that this makes sense, I will update this PR and hold it until the next release.

tashian:
This is complete, but it's blocked on new open source releases.

dopey (approved):

PR #291, opened 2023-12-04
Update Intune tutorial to fix UX gap

tashian:


Linda-Ikechukwu (commented):
This looks good. Two things I noticed are (couldn't comment directly on the specific lines): 

Step three, ### 3. Configure Smallstep needs to be changed to H2 to match other headings. 

For separation of concerns and SEO reasons, and easier findability, I think it would be ideal to move the section on "Adding Wi-Fi Support" to a new page, and then link to where we have the phrase "This tutorial assumes you are configuring EAP-TLS Wi-Fi network access." or just link to it

tashian:
@Linda-Ikechukwu I've updated the tutorial based on your feedback, thanks. Could you give it another review? I didn't separate out the Wi-Fi bits yet, but it's a good idea for the future. For now I just want to get this out there. All of this is changing quickly and will continue to evolve as we automate more bits of the process.

Linda-Ikechukwu (approved):

PR #299, opened 2024-02-07
Add ACME Wire configuration

hslatman:


step-ca/provisioners.mdx
```diff
@@ -1027,6 +1030,91 @@ To get a client certificate for a hardware-bound private key on your YubiKey:
...
+          "dpop": {
+            "key": "base64(public.pem)",
+            "target": "https://example.com/clients/{{ .DeviceID }}/access-token"
+          }
+      }
+    }
+}
+```
+
+- The `wire-oidc-01` and `wire-dpop-01` challenge types must both be enabled.
```
tashian: A small formatting request. I think this bulleted list should align with same style we use for other other provisioner config descriptions. See [JWK](https://github.com/smallstep/docs/blob/main/step-ca/provisioners.mdx#L353) for example. The field name in bold, followed by a description of the available options and any other operational notes related to how this field interacts with this provisioner.

step-ca/provisioners.mdx
```diff
@@ -1027,6 +1030,91 @@ To get a client certificate for a hardware-bound private key on your YubiKey:
...
+#### ACME for Wire messenger clients
+
+<Alert severity="warning">
+  <div> 
+    The challenge types for Wire clients must be explicitly enabled in an ACME provisioner.
+    Currently the configuration can only be done through editing the configuration file.
+  </div>
+</Alert>
+
+The `wire-oidc-01` and `wire-dpop-01` challenges enable `step-ca` to issue certificates to Wire clients.
```
tashian:
```suggestion
The `wire-oidc-01` and `wire-dpop-01` challenges enable `step-ca` to issue certificates to [Wire](https://wire.com) messenger clients.
```

step-ca/provisioners.mdx
```diff
@@ -1027,6 +1030,91 @@ To get a client certificate for a hardware-bound private key on your YubiKey:
...
+Wire clients make use of a user and a device identity, both of which are challenged and verified, after which the identities are combined into a single X.509 certificate.
```
tashian, line 1043: We have both broad (eg. OIDC) and narrow (eg. AWS) provisioner types. Where does Wire fit in? Could these new Wire challenge types offer a combined device + user identity certificate flow for any application? How useful are they outside of Wire clients? Would a user have to write their own ACME client from scratch for these challenge types, or is there a reference implementation available in open source?

If the Wire challenge types have a more broad potential, do we have the right names for them? And for this section of the docs?

tashian (changes requested):

step-ca/provisioners.mdx
hslatman: Hehe, I only looked at the SCEP one, which has this style. I'll update this; and likely also the SCEP one, if it makes sense.

step-ca/provisioners.mdx
hslatman: > We have both broad (eg. OIDC) and narrow (eg. AWS) provisioner types. Where does Wire fit in? 

The Wire integration uses ACME as the core protocol, with two new identifier types, and two new challenge types. It's a subtype of the ACME provisioner, similar to ACME for Device Attestation, so that's why it's on the same level as that. It uses both OIDC and DPoP, so there are some similarities to the OIDC provisioner, but it should be considered a separate thing.

> Could these new Wire challenge types offer a combined device + user identity certificate flow for any application? How useful are they outside of Wire clients? 

Potentially yes, and we've looked at doing so. There's some overlap with existing draft RFCs for extensions to ACME, as well as some other related ones. The verdict was that it would most probably take quite long to get to a place where we would have a "fully standard approach", so that's why this extension is now fairly tightly coupled to just Wire. We decided to put configuration in a new `wire` object, apart from enabling the challenge types, containing the configuration neatly together. This should make refactoring easier, if we were to reconsider the decision to make this Wire specific. That said, similar concepts as used in this integration would be nice candidates to generalize, and we can learn from this integration being used in practice to see what makes sense. Besides that, I think Wire would start the process of putting these new identifiers and challenges in the official registries.

> Would a user have to write their own ACME client from scratch for these challenge types, or is there a reference implementation available in open source?

Wire is actively working on integrating this in their clients. I don't know the exact status, but I think the [web client](https://github.com/wireapp/wire-webapp) implementation is furthest along. They have an end-to-end testing framework that also performs the full ACME flow using the new identifier types written in Rust as part of [rusty-jwt-tools](https://github.com/wireapp/rusty-jwt-tools). Maybe a core Rust library supporting these is an option. Outside of the Wire ecosystem, users would be pretty much on their own.

> If the Wire challenge types have a more broad potential, do we have the right names for them? And for this section of the docs?

We decided to prefix everything with `wire-` now, so that things are namespaced and should be less of an issue if we were to generalize them.

step-ca/provisioners.mdx
hslatman: I've updated the style. I decided to not include all the new layers, and to only put the leaf level as the new bullets to keep it simple. Alternatively, could use dotted levels?

SCEP changes are in #303

step-ca/provisioners.mdx
rohan-wire: Just adding onto to what Herman said:

Regarding stable documentation of the ACME extensions in a neutral location, I'll have an internet draft by 4-Mar-2024 (draft-mahy-acme-wire-types), and I'll inform the ACME WG mailing list and request publication of an individual RFC.

Regarding ACME client implementations, all the Wire clients (web, iOS, Android) have implemented the full ACME flow with these identifiers and challenges. All this client code and the Wire crypto libraries are open source.

step-ca/provisioners.mdx
```diff
@@ -1027,6 +1030,96 @@ To get a client certificate for a hardware-bound private key on your YubiKey:
...
+Here's an example of an ACME provisioner with Wire support enabled in `$(step path)/config/ca.json`:
+
+```json
+...
+{
+    "type": "ACME",
+    "name": "wire",
+    "claims": {
+        "maxTLSCertDuration": "24h",
+        "defaultTLSCertDuration": "16h"
```
rohan-wire: Wire identity certs are necessarily long lived because a received message might be decoded up to 28 days later. Wire recommends a Wire identity certificate is valid for 60-90 days. 

```suggestion
        "maxTLSCertDuration": "2160h",
        "defaultTLSCertDuration": "1440h"
```

step-ca/provisioners.mdx
```diff
@@ -1027,6 +1030,96 @@ To get a client certificate for a hardware-bound private key on your YubiKey:
...
+    "challenges": [
+        "wire-oidc-01",
+        "wire-dpop-01"
+    ],
+    "options": {
+      "x509": {
+        "template": 
+          "{
+            \"subject\": {
+                \"organization\": \"Organization Name\",
```
rohan-wire: We recommend using the domain name of the organization to identify it uniquely.

```suggestion
                \"organization\": \"example.com\",
```

step-ca/provisioners.mdx
```diff
@@ -1027,6 +1030,96 @@ To get a client certificate for a hardware-bound private key on your YubiKey:
...
+                    "ES256",
+                    "ES384",
+                    "EdDSA"
+                ]
+            },
+            "transform": "{\"name\": \"{{ .name }}\", \"preferred_username\": \"{{ .preferred_username }}\"}"
+          },
+          "dpop": {
+            "key": "base64(public.pem)",
+            "target": "https://example.com/clients/{{ .DeviceID }}/access-token"
```
rohan-wire:
```suggestion
            "target": "https://wire-server.example.com/clients/{{ .DeviceID }}/access-token"
```

step-ca/provisioners.mdx
```diff
@@ -1027,6 +1030,96 @@ To get a client certificate for a hardware-bound private key on your YubiKey:
...
+* **jwksUrl**<Reference id="star6" marker="*" />: sets the OIDC JWKS URL. Ignored if `discoveryBaseUrl` is set.
+* **userInfoUrl**<Reference id="star6" marker="*" />: sets the OIDC userInfo URL. Ignored if `discoveryBaseUrl` is set.
+* **provider.signatureAlgorithms**<Reference id="star6" marker="*" />: sets the supported OIDC signature algorithms. Ignored if `discoveryBaseUrl` is set.
+* **clientId**: sets the OIDC client ID used in OIDC token validation.
+* **config.signatureAlgorithms**<Reference id="star6" marker="*" />: sets the OIDC signature algorithms to be used in validation.
+* **transform**<Reference id="star6" marker="*" />: a Go template that can be used to map and/or transform properties in the OIDC token. Different SSO IdPs include different claims in their OIDC tokens. Using this `transform` template, the tokens are verified first, then transformed and persisted. Transformation uses the same engine as our certificate template engine. The properties from the transformed token can be referenced in the certificate template configured in the `x509` option.
```
rohan-wire: Please consider adding a link to the Go template syntax.

step-ca/provisioners.mdx
```diff
@@ -1027,6 +1030,96 @@ To get a client certificate for a hardware-bound private key on your YubiKey:
...
+* **provider.signatureAlgorithms**<Reference id="star6" marker="*" />: sets the supported OIDC signature algorithms. Ignored if `discoveryBaseUrl` is set.
+* **clientId**: sets the OIDC client ID used in OIDC token validation.
+* **config.signatureAlgorithms**<Reference id="star6" marker="*" />: sets the OIDC signature algorithms to be used in validation.
+* **transform**<Reference id="star6" marker="*" />: a Go template that can be used to map and/or transform properties in the OIDC token. Different SSO IdPs include different claims in their OIDC tokens. Using this `transform` template, the tokens are verified first, then transformed and persisted. Transformation uses the same engine as our certificate template engine. The properties from the transformed token can be referenced in the certificate template configured in the `x509` option. 
+* **key**: the base64 encoded PEM public key corresponding to the private key that signs the DPoP token.
+* **target**: a templated URL pointing to the token issuer. The value of `DeviceID` is evaluated when the certificate is requested.
```
rohan-wire:
```suggestion
* **target**: a templated URL pointing to the token issuer (the Wire server). The value of `DeviceID` is evaluated when the certificate is requested.
```

rohan-wire (changes requested):
I included a few small suggestions. Thanks!

step-ca/provisioners.mdx
hslatman: Added in 4c42e0dc5faaf8b42fd9c8e26e7bf7571f1a52b1.

rohan-wire (approved):

hslatman:
@tashian the SCEP changes got merged into this one. Can you re-review the latest changes in this one?

tashian (approved):

PR #353, opened 2024-10-18
templates: Fix unicode symbols typo in command example

nh2:
Simple typo fix.


tashian:
Fixed in a different PR :)

PR #366, opened 2024-12-18
Add note about device identity

tashian:


step-ca/provisioners.mdx
```diff
@@ -1144,9 +1152,6 @@ TPMs also have privacy constraints and a wide range of use case-specific validat
   </div>
 </Alert>
 
-For a description of how a TPM-based attestation flow might work, if implemented,
-see [TPMs + ACME: Better Together](https://smallstep.com/blog/managed-device-attestation/#tpms--acme-better-together)
-
```
hslatman, line 1149: Is your reason for deleting this because it shows how things work, or just that it's in the wrong place? I think the diagram and text are at a fairly high level, so not too big of an issue?

hslatman (approved):
See question

PR #396, opened 2025-03-25
Some clarifications in the Templates doc

tashian:
Just revisiting some vocabulary. I realized that template "variables" are actually constants, so I wanted to clarify that.
Also updated some language around `.Insecure` values.


step-ca/templates.mdx
```diff
@@ -41,9 +42,9 @@ A few custom [functions for ASN.1 encoding](#asn1-values) and [time formatting](
 <Alert severity="warning">
   <div>
     <strong>Warning: Always wrap values in <code>toJson</code></strong><br />
-    In the templates on this page, variables are pulled into templates using <code>{`{{`} toJson .variableName {`}}`}</code> 
-    to sanitize the value of the variable.
-    When using templates, you must sanitize all variables using <code>toJson</code> to avoid template injection vulnerabilities.
+    In the templates on this page, constants are pulled into templates using <code>{`{{`} toJson .constantName {`}}`}</code>
```
hslatman: Not sure if constant is the right word to use instead of variables. 

In the Go template docs they're referred to as fields and keys, depending on whether it's a struct or a map that values are being referenced from. The value that the field/key points to is not constant; that is a variable value (and I think that's why they were called variables in the first place). Some alternatives, depending on the context:

- Named values
- Named properties
- Fields
- Keys

Later on you use values. I think that makes sense in that context, as it discusses the dynamic values that are referenced by the properties.

Some of the keys are defined by us in default templates and other places in the code. Those names are indeed "constant", but in a different way. More like static, or predefined named values.

tashian: @hslatman Ah, I see. So, what happens if I change a variable like .SANs in the template? Is there any situation in which it would be desirable to change one of these variables?

hslatman: You would generally only "change" them to not be used, or to do a certain transformation, such as picking the first of `.SANs`, or do a substring transformation; things like that. 

However, that means it is changing the value that `.SANs` points to; not the `.SANs` property name itself. We control which properties are available in the template (for the most part; `.Insecure.User` can have user-defined property names, for example). 

I can understand that that can be confusing: the `.SANs` key (and other keys) are "constant", in that they're defined by us, and set to a value in a controlled manner. But the values they're pointing to are _not_ constant, which is what I was trying to say with my previous comment. 

For example:

`In the templates on this page, constants are pulled into templates ...`

I wouldn't call that "pulling a constant into a template". It's pulling the value from the field identified by name (i.e. `SANs`) into the template.

tashian: I'll switch it back to variables then, both because it's literally true, but also because it's colloquially a fine word choice.

hslatman (approved):

PR #418, opened 2025-06-18
[CON-232] Update Intune docs for MSI installer

tashian:
I uploaded an MSI into Intune, but need to get confirmation that this actually works:
- [x] Any other fields we need to configure in the App information pane?
- [ ] Does it install and run properly?
- [x] How does Intune differentiate arm64 and amd64? Do we need special managed app filters, or does it work to just add both installers? The managed app filtering options don't seem to include filtering by architecture.

### DIFF
--- a/platform/core-concepts.mdx
+++ b/platform/core-concepts.mdx
@@ -1,91 +1,346 @@
 ---
+updated_at: July 07, 2025
 title: Core Concepts
 html_title: Smallstep Core Concepts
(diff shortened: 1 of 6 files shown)


hslatman:
Simplifies the setup a lot 😄

hslatman:
@alanchrt fwiw: I find using magic Linear keywords to be more flexible and powerful than branch naming and/or issue title. It also keeps this type more like metadata, instead of polluting Git with ephemeral and (when used out of context), unrelated identifiers in Git message titles.

alanchrt (approved):

PR #436, opened 2025-09-18
Add section on restricting Jamf re-deployment of identical version

dopey:


tashian:
Reading what you wrote, I'm still a little confused.

Could they alternatively use a different [execution frequency](https://learn.jamf.com/en-US/bundle/jamf-pro-documentation-current/page/Execution_Frequency_for_Policies.html)? Right now we suggest a frequency of Ongoing. What's wrong with using Once per week or Once per day?

Also, right now we instruct them to trigger the policy at Login. Does the Login trigger also cause a policy to run on a cadence? In my mental model, Login + Ongoing means the policy will only run each time I login.

dopey:
> Could they alternatively use a different [execution frequency](https://learn.jamf.com/en-US/bundle/jamf-pro-documentation-current/page/Execution_Frequency_for_Policies.html)? 

Sure. I think we spoke about this yesterday - ultimately these decisions are up to the IT admins. We're just there to recommend something sane. 

> Right now we suggest a frequency of Ongoing. What's wrong with using Once per week or Once per day?

There's nothing inherently wrong with other frequencies. Not really up to us, how IT admins want to manage the software on their fleets. All we're doing is making a recommendation. And, to my understanding, the recommendation of using the Smart Computer Group to scope distribution is still valid, regardless of the configured frequency.

tashian (approved):

PR #439, opened 2025-10-01
[CON-265] Add ChromeOS to basic platform lists

tashian:


platform/README.mdx
```diff
@@ -56,7 +56,7 @@ This process, known as cryptographic device attestation, forms the foundation fo
 
 # How can you use Smallstep?
 
-The Smallstep Agent is the vehicle through which Smallstep delivers cryptographically attested device identity to your organisation. It is the recommended way to identify devices and get client certificates to devices (Windows, Linux, Mac OS) for Enterprise Wi-Fi, VPN, HTTP/3 proxies, or web applications. 
+The Smallstep Agent is the vehicle through which Smallstep delivers cryptographically attested device identity to your organisation. It is the recommended way to identify devices and get client certificates to devices (Windows, Linux, Mac OS, ChromeOS) for Enterprise Wi-Fi, VPN, HTTP/3 proxies, or web applications.
```
hslatman:
```suggestion
The Smallstep Agent is the vehicle through which Smallstep delivers cryptographically attested device identity to your organisation. It is the recommended way to identify devices and get client certificates to devices (Windows, Linux, macOS, ChromeOS) for Enterprise Wi-Fi, VPN, HTTP/3 proxies, or web applications. 

```

hslatman: Also curious why it's specifically HTTP/3. Just because it's novel? Or as a general term for network relay, which happens to be HTTP/3 (generally, but can be HTTP/2).

platform/README.mdx
```diff
@@ -68,7 +68,7 @@ If for any reason, you cannot have the Smallstep Agent on your devices, Smallste
 
 Smallstep integrates with your MDM to deploy client certificates to company-managed devices to enable certificate-based network authentication for Wi-Fi (802.1x EAP-TLS WPA-Enterprise), VPN, ZTNA, etc.
 
-We offer integrations for any MDMs for Apple and Windows devices that support Dynamic SCEP like Jamf, Intune, Workspace ONE, Mosyle, Ivanti, e.t.c.
+We offer integrations for any MDMs for Apple, Windows, and ChromeOS devices that support Dynamic SCEP like Jamf, Intune, Workspace ONE, Mosyle, Ivanti, and Google Workspace.
```
hslatman, line 72: This is in the context of dynamic SCEP. Technically, we can do it and have tested it, but I don't think the current implementation is something we want to support officially. We would likely want to iterate some more on it properly, if users need it. Functionally, there's no need for SCEP on ChromeOS within our platform.

platform/README.mdx
tashian: I think we were using "HTTP/3 proxy" as a shorthand for MASQUE?

andysmallstep (approved):
Sections looked like logical improvements ot the textual contents. lgtm

PR #461, opened 2025-11-18
Workspace ONE script assignment

joshdrake:
Updates to correct workflow in WS1.

Thank you!


tashian (approved):
This looks good. @joshdrake I added the `capi:` uri change to the Intune doc as well

PR #473, opened 2025-12-15
fix: remove duplicate heading

C-Mejlak:
The `<h1>` is automatically grabbed from the `front_matter` or the `title` defined in `manifest.json`.

This is resulting in a duplication of the heading.

image

Thank you!


andysmallstep (approved):
Easy enough

PR #507, opened 2026-04-29
Document winget install for the Smallstep Agent on Windows

tashian:
Smallstep.step-agent is now published to microsoft/winget-pkgs, so add it to the Windows installation section as the recommended quick path above the manual MSI download.

Refs EFF-174.

Thank you!


andysmallstep (approved):
I also tested this command set on Windows 11 Pro, installed with no error.

PR #516, opened 2026-05-21
Document Windows per-user reloader failure mode

joshdrake:
## Summary

Adds Windows-specific content to `platform/troubleshooting-agent.mdx` that the page is currently missing:

- A short **Windows agent architecture** subsection under `Platform-specific reference > Windows` so readers understand the two-process model (system service + per-user scheduled task + reloader pipe).
- A new **Certificate storage fails with "cannot find the file specified" (Windows)** entry under `Common issues and solutions`, including the exact PowerShell / `schtasks` commands and recovery steps.
- Cross-links from the existing **Agent won't start** and **Certificate expired or invalid** entries so a reader who lands there with these symptoms can find the right page.

The new troubleshooting entry covers both error variants:
- the recent wrapped agent error introduced in [smallstep/agent#1052](https://github.com/smallstep/agent/pull/1052)
- the older raw gRPC dial error that users on earlier agent versions will see

## Test plan

- [ ] Render locally with `pnpm dev` in the smallstep.com repo and verify the new headings, anchors, and cross-links.
- [ ] Confirm a teammate can walk from the raw error string to a recovery action using only the docs.

### DIFF
--- a/platform/troubleshooting-agent.mdx
+++ b/platform/troubleshooting-agent.mdx
@@ -407,6 +407,10 @@ This outputs check results in JSON format:
 4. Restart the system if the TPM is in an inconsistent state
 5. On Linux, check for conflicting services using the same ports
 
+<Alert severity="info">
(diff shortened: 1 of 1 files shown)


tashian (approved):
This page desperately needs a table of contents, and our TOC sidenav is not visible unless the browser is full width on desktop. That's a structural issue beyond the scope of this PR. Looks good

PR #526, opened 2026-06-15
[PRE-91] Use downloadable config profile for Jamf/Fleet

JoeWagstaff:
#### Describe your changes:
Jamf: Replace the manual "Configure an agent enrollment profile" steps (hand-built Managed Login Items / Certificate / SCEP / External Applications payloads) with downloading the pre-built profile from the dashboard (Device Management → connection → Settings → Configuration Profile (macOS only)) and uploading it to Jamf; scope it to the agent install policy.
Fleet: Replace the hand-authored smallstep-agent.mobileconfig (Step 4 XML template + base64 CA paste) with the downloaded fleet-<connection-id>.mobileconfig; update the upload step and the GitOps section to reference it.
Fleet: Tighten Step 3 to require naming the CA SMALLSTEP_AGENT (the generated profile hardcodes that suffix); drop the stray $FLEET_VAR_HOST_END_USER_EMAIL_IDP variable (not used by the profile).
Add notes clarifying the profiles rely on Fleet/Jamf substituting their dynamic SCEP values at deploy time, and that the macOS-only profile's SCEP payload still serves iOS/iPadOS.
Keep the API-client/connection/webhook setup steps (the download depends on them).

#### Related links/other PRs/issues:

Thank you!

PR #533, opened 2026-07-21
Update Jamf and macOS w/ Fleet and agent documentation

hslatman:
This PR takes the changes from #526, and builds on it by replacing the Fleet SCEP enrollment flow with an ACME enrollment flow backed by our Osquery extension. At the moment the `.mobileconfig` in our web console has to be tweaked manually for this flow to work: 1) `Certificate` to be removed; 2) `CAFingerprint` to be set to the SHA256 of the agents root.

Some iOS/ipadOS/tvOS references have been removed, as they seemed to be out of place in the current document. For those platforms we should either support plain SCEP (without the agent) or native ACME. It warrants an entirely new section or possibly even a page.

The updated Jamf docs have not been changed.


tutorials/connect-fleet-dm-to-smallstep.mdx
```diff
@@ -92,204 +95,69 @@ The default role for API-only users is **Observer**, which grants read-only acce
...
+## Step 4. Download the configuration profile
 
-Fleet will test the CA connection after you create it.
-
-<Alert severity="info">
-<div>
-If you plan to use GitOps instead of the Fleet UI, skip this step and see the [GitOps section below](#gitops-configure-fleet-with-fleetctl) for the YAML-based alternative.
-</div>
-</Alert>
+Fleet deploys certificates and configuration to devices using a configuration profile. You can download this from the Smallstep dashboard and upload it to Fleet. No need to build the payloads or paste certificate data by hand.
```
J-Hunter-Hawke, line 141: The PR description notes that the `.mobileconfig` from the web console currently still has to be tweaked manually for this flow to work — specifically (1) removing the `Certificate` key and (2) setting `CAFingerprint` to the SHA256 of the agent's root. That seems to contradict the "No need to build the payloads or paste certificate data by hand" instruction here (and the "bundles everything required" follow-up on line 147).

Should these steps document the manual `Certificate` removal and `CAFingerprint` change, or is the console expected to emit a ready-to-upload profile before this merges? Want to make sure we're not shipping instructions that won't work end-to-end yet.

(Note: the Jamf doc has the equivalent "no need to build the payloads by hand" claim on its line 136 — may warrant the same caveat if it applies there too.)

hslatman: The console is expected to fill the correct info. That change won't be online before this gets tested, though. So just manual for now, and this can be merged after.

For Jamf there's no need to change it, as it still uses SCEP in that case.

J-Hunter-Hawke: >1) Certificate to be removed; 2) CAFingerprint to be set to the SHA256 of the agents root.

For personal clarity, this is referring to the mcx_preference_settings that the agent uses; it is preconfigured to look for the SCEP certificate

PR #545, opened 2026-08-01
Add ChromeOS device identity certificates tutorial

llewis1234:
## Summary
- New tutorial covering issuing mTLS-capable device identity certificates to ChromeOS devices (Devices/Accounts authority, ACME Device Attestation, trust distribution, forced re-enrollment, verification, troubleshooting)
- Extends the Google Workspace integration guide with the Chrome Verified Access API setup, `verifiedaccess` OAuth scope, Verified Access Mode config, and ChromeOS extension deployment steps that the new tutorial depends on
- Cross-links from the agent troubleshooting doc and adds the new tutorial to the nav manifest

## Test plan
- [ ] `vale` run against changed files (noise-filtered against house style — no unaddressed findings)
- [ ] `markdown-link-check` run against changed files — all internal/external links resolve
- [ ] Visual preview via the docs renderer (not yet done — recommend before merge)

### DIFF
--- a/tutorials/chromeos-device-identity-certificates.mdx
+++ b/tutorials/chromeos-device-identity-certificates.mdx
@@ -1,0 +1,105 @@
+---
+updated_at: July 31, 2026
+title: Configure ChromeOS Device Identity Certificates
+html_title: Configure ChromeOS Device Identity Certificates for mTLS | Smallstep
+description: Issue mTLS-capable device identity certificates to ChromeOS devices enrolled in Google Workspace, for use in Wi-Fi, VPN, or internal application authentication.
+---
+
+Smallstep issues two different kinds of certificate to a ChromeOS device enrolled through [Google Workspace](./connect-google-workspace-to-smallstep.mdx): a short-lived **attestation certificate**, issued automatically, and a longer-lived **device identity certificate**, which is what you actually use for mTLS — Wi-Fi (802.1X), VPN, or authenticating to internal applications.
+
+This guide covers issuing that device identity certificate. It's a prerequisite for [ChromeOS Wi-Fi Enrollment](./protect-wireless-networks.mdx#chromeos-with-google-workspace-acme-device-attestation) and for any other mTLS use case involving ChromeOS devices.
+
+# Prerequisites
+
+- A completed [Connect Google Workspace to Smallstep](./connect-google-workspace-to-smallstep.mdx) integration, with devices syncing into Smallstep
+- The Smallstep ChromeOS extension deployed to your target Organizational Unit (covered in that same guide)
+
+# Attestation certificate vs. device identity certificate
+
+A freshly enrolled ChromeOS device receives only the attestation certificate. It cannot be used for mTLS. Until you complete the steps below, the device will not have a usable identity certificate, even though it will appear fully enrolled and "high assurance" in both the Smallstep dashboard and the Admin Console.
+
+| | Attestation certificate | Device identity certificate |
+|---|---|---|
+| Issued by | Team's Intermediate CA | Devices (or Accounts) Intermediate CA |
+| Lifetime | 24 hours | 30 days |
+| Subject alternative name | `urn:ak:sha256:…` (attestation key) | `urn:smallstep:device:permanent-identifier:<serial>` |
+| Purpose | Proves genuine hardware after Chrome Verified Access; used as the credential in the ACME Device Attestation exchange | The credential presented for mTLS |
+| Available from Certificate Manager? | No — not published to customers | Yes |
+
+<Alert severity="info">
+<div>
+Only the device identity certificate can be used for mTLS. If a device presents only the attestation certificate to an mTLS endpoint, the connection will fail — see Troubleshooting below.
+</div>
+</Alert>
+
+# Step-by-step instructions
+
+### 1. Choose an issuing authority
+
+The device identity certificate can be issued from either of two authorities in Certificate Manager:
+
+- **Devices** — binds the certificate to the device itself (subject is the hardware serial number). Use this for device-level authentication such as Wi-Fi.
+- **Accounts** — binds the certificate to the Google Workspace user currently signed in to the device (subject is the user's email). Use this if you need to authenticate the user rather than the device.
+
+Both authorities support ACME Device Attestation, so the enrollment mechanics in the steps below are identical either way — only the resulting certificate's identity differs.
+
+### 2. Distribute the authority's trust root to Google Admin
+
+This step must happen **before** step 3, not after. Each authority's ACME directory is served over TLS using a certificate signed by that authority's own private root. If the Chromebook doesn't already trust that root, it cannot reach the directory at all — and the failure is completely silent. No error appears in the extension, in `chrome://policy`, or in the Smallstep dashboard; the device simply keeps its existing certificate indefinitely.
+
+1. In Certificate Manager, go to **Authorities**, select the authority you chose in Step 1, and download its **Root Certificate** and **Intermediate Certificate**.
+2. In Google Admin, go to **Devices → Networks → Certificates**, select your target Organizational Unit, and add both certificates. Check ✅ **Enabled for Chromebook** for each.
+
+<Alert severity="warning">
+<div>
+You can confirm whether trust distribution succeeded from any machine, before touching a Chromebook: `curl -sS https://<authority>.<team-slug>.ca.smallstep.com/acme/acme-da/directory`. A TLS error like `unable to get local issuer certificate` means the root hasn't propagated yet or wasn't uploaded. A JSON response means you're clear to continue.
+</div>
+</Alert>
+
+### 3. Set the extension policy
+
+In Google Admin, go to **Devices → Chrome → Apps & Extensions → Users & browsers**, select your Organizational Unit, and open the Smallstep extension's **Policy for extensions**. Add `devices_ca_directory_url`, pointing at the ACME Device Attestation directory of the authority you chose:
+
+```json
+{
+  "team": { "Value": "<team-slug>" },
+  "devices_ca_directory_url": {
+    "Value": "https://devices.<team-slug>.ca.smallstep.com/acme/acme-da/directory"
+  }
+}
+```
+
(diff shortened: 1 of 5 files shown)


llewis1234:
Follow-up ready and waiting: #546 (ChromeOS Wi-Fi enrollment via ACME Device Attestation) is stacked on this branch and opened as a draft. It's blocked on this PR merging — once this lands, #546's diff will collapse to just its own changes and it'll be ready for review.

tutorials/chromeos-device-identity-certificates.mdx
```diff
@@ -0,0 +1,105 @@
+---
+updated_at: July 31, 2026
+title: Configure ChromeOS Device Identity Certificates
+html_title: Configure ChromeOS Device Identity Certificates for mTLS | Smallstep
+description: Issue mTLS-capable device identity certificates to ChromeOS devices enrolled in Google Workspace, for use in Wi-Fi, VPN, or internal application authentication.
+---
+
+Smallstep issues two different kinds of certificate to a ChromeOS device enrolled through [Google Workspace](./connect-google-workspace-to-smallstep.mdx): a short-lived **attestation certificate**, issued automatically, and a longer-lived **device identity certificate**, which is what you actually use for mTLS — Wi-Fi (802.1X), VPN, or authenticating to internal applications.
```
hslatman: I think the fact that there exists an attestation certificate can be removed. It's true that it gets issued, and it will show up in the ChromeOS certificate store, but from the user's perspective it's an implementation detail that they don't necesssarily need to know about in daily usage. It's more something to explain in a more technical "how does it work" type of article.

Other mentions of the attestation certificate should also be adapted.

May need discussion with product on whether or not to remove it.

tutorials/chromeos-device-identity-certificates.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+A freshly enrolled ChromeOS device receives only the attestation certificate. It cannot be used for mTLS. Until you complete the steps below, the device will not have a usable identity certificate, even though it will appear fully enrolled and "high assurance" in both the Smallstep dashboard and the Admin Console.
+
+| | Attestation certificate | Device identity certificate |
+|---|---|---|
+| Issued by | Team's Intermediate CA | Devices (or Accounts) Intermediate CA |
+| Lifetime | 24 hours | 30 days |
+| Subject alternative name | `urn:ak:sha256:…` (attestation key) | `urn:smallstep:device:permanent-identifier:<serial>` |
+| Purpose | Proves genuine hardware after Chrome Verified Access; used as the credential in the ACME Device Attestation exchange | The credential presented for mTLS |
+| Available from Certificate Manager? | No — not published to customers | Yes |
```
hslatman: "Team's Intermediate CA" is too generic. It comes from the attestation CA, and is signed by its intermediate.

But, taking into account the earlier remark, this table may not be necessary in this tutorial.

tutorials/chromeos-device-identity-certificates.mdx
```diff
@@ -0,0 +1,105 @@
...
+<div>
+Only the device identity certificate can be used for mTLS. If a device presents only the attestation certificate to an mTLS endpoint, the connection will fail — see Troubleshooting below.
+</div>
+</Alert>
+
+# Step-by-step instructions
+
+### 1. Choose an issuing authority
+
+The device identity certificate can be issued from either of two authorities in Certificate Manager:
```
hslatman: Technically it can come from more authorities. Decision to be made whether or not we want to expose these options to users. So far we've been in the loop when setting up configurations with our customers, because we don't have good UX for our users to do it fully on their own.

tutorials/chromeos-device-identity-certificates.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+- **Devices** — binds the certificate to the device itself (subject is the hardware serial number). Use this for device-level authentication such as Wi-Fi.
+- **Accounts** — binds the certificate to the Google Workspace user currently signed in to the device (subject is the user's email). Use this if you need to authenticate the user rather than the device.
+
+Both authorities support ACME Device Attestation, so the enrollment mechanics in the steps below are identical either way — only the resulting certificate's identity differs.
+
+### 2. Distribute the authority's trust root to Google Admin
+
+This step must happen **before** step 3, not after. Each authority's ACME directory is served over TLS using a certificate signed by that authority's own private root. If the Chromebook doesn't already trust that root, it cannot reach the directory at all — and the failure is completely silent. No error appears in the extension, in `chrome://policy`, or in the Smallstep dashboard; the device simply keeps its existing certificate indefinitely.
```
hslatman: The extension will actually keep retrying in the background if it can't connect. If it doesn't, let me know, as that would be a bug.

We chose to keep it silent on the client side, since users on the device themselves can't do much themselves to fix things, except for checking with their IT department. In the future we'll likely have the ChromeOS extension send telemetry with which we can inform IT admins something's wrong.

tutorials/chromeos-device-identity-certificates.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+1. In Certificate Manager, go to **Authorities**, select the authority you chose in Step 1, and download its **Root Certificate** and **Intermediate Certificate**.
+2. In Google Admin, go to **Devices → Networks → Certificates**, select your target Organizational Unit, and add both certificates. Check ✅ **Enabled for Chromebook** for each.
+
+<Alert severity="warning">
+<div>
+You can confirm whether trust distribution succeeded from any machine, before touching a Chromebook: `curl -sS https://<authority>.<team-slug>.ca.smallstep.com/acme/acme-da/directory`. A TLS error like `unable to get local issuer certificate` means the root hasn't propagated yet or wasn't uploaded. A JSON response means you're clear to continue.
+</div>
+</Alert>
```
hslatman, line 43: This sounds more like an `info`.

The audience for these docs are IT admins responsible for managing a fleet of ChromeOS machines. They don't have direct access to the machines, unless they access the machine using remote desktop? I also think `curl` requires the terminal to be available, which I think is only the case when the Linux development environment is available, which may not always be the case by default.

tutorials/chromeos-device-identity-certificates.mdx
```diff
@@ -0,0 +1,105 @@
...
+```json
+{
+  "team": { "Value": "<team-slug>" },
+  "devices_ca_directory_url": {
+    "Value": "https://devices.<team-slug>.ca.smallstep.com/acme/acme-da/directory"
+  }
+}
+```
+
+Use `devices.<team-slug>...` for the Devices authority, or `accounts.<team-slug>...` for the Accounts authority.
```
hslatman: At the moment, `devices.` is the default authority that will be configured. The ChromeOS extension should be getting the device identity certificate by default, without having to specify `devices_ca_directory_url`; it should need just the `team"`. Consider rewriting this a bit to say that it's only needed when a different cert has to be obtained.

If the extension doesn't get the device identity cert by default (assuming root is in place), let me know, as that would be a bug.

tutorials/chromeos-device-identity-certificates.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+### 4. Force a fresh enrollment
+
+The extension will not re-enroll while it already holds a valid certificate, so a policy save alone has no effect. On the Chromebook:
+
```
hslatman, line 65: Did you have a specific use case for wanting to trigger this manually?

tutorials/chromeos-device-identity-certificates.mdx
```diff
@@ -0,0 +1,105 @@
...
+The reboot matters — a device can otherwise continue presenting its old certificate to an mTLS endpoint even after the new one has been issued.
+
+### 5. Verify certificate issuance
+
+On the Chromebook, go to `chrome://certificate-manager/clientcerts/platformclientcerts`. You should see two certificates: the original attestation certificate, and a new one issued by your chosen authority's Intermediate CA. Open the new certificate's **Details** and confirm the **Issuer** matches the authority you configured in Step 1.
+
+<Alert severity="info">
+<div>
+The Smallstep dashboard's device detail page may continue to show "This device does not have any identity certificates" even after a certificate has been issued and is actively in use. ChromeOS certificates are issued over SCEP/ACME rather than the agent path this panel tracks, so it isn't a reliable place to check. Confirm issuance on the device itself, or under **Certificate Manager → Certificates → x509** in the Smallstep dashboard, where issued certificates are listed normally.
```
hslatman: The device identity not showing up is actually a deficit in the ChromeOS extension at the moment. We actually intend the ChromeOS extension to act more like the full agent, and thus its device identity certificate should eventually show up in our platform. 

The fact that it was issued through ACME (we don't use SCEP on ChromeOS) isn't the issue here.

tutorials/connect-google-workspace-to-smallstep.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-updated_at: February 03, 2026
+updated_at: August 01, 2026
```
hslatman: As mentioned in Slack, let's try to rebase this on top of my latest changes or after that gets merged.

llewis1234:
All inline feedback addressed:

- Attestation certificate removed as a user-facing concept (L8/L27) — it's now just "the certificate," no comparison table
- Step 1 reframed from "choose an authority" to "confirm your issuing authority," since this is normally set up together with a Smallstep contact rather than self-service (L39)
- Corrected the trust-distribution failure description — the extension does retry in the background, it just doesn't surface that to the admin (L48)
- Downgraded the curl verification tip to `info` severity, clarified it runs from your own machine, not the Chromebook (L43)
- Clarified `devices_ca_directory_url` is only needed for the Accounts authority or a non-default Devices setup — not required by default (L72)
- Fixed a related bug this surfaced: the default-authority path was pointing readers to skip to the wrong step (Step 4 instead of Step 5)
- Clarified Step 4 (force re-enrollment) only applies when Step 3's policy actually changed on an already-enrolled device — never on first-time setup, which answers the "specific use case" question (L65)
- Corrected the dashboard-gap explanation — it's a known gap in the ChromeOS extension today, unrelated to ACME vs. SCEP (L91)
- Rebased onto `herman/chromeos-wif` per our Slack conversation (L2)

Ready for review whenever you get a chance.

tashian (approved):

PR #549, opened 2026-08-26
NixOS: agent.yaml can be declared, not registered

joshdrake:
Refs [OFF-19](https://linear.app/smallstep/issue/OFF-19/step-agentservice-gate-on-config-existence-not-filesystem-writability).

[smallstep/agent#1205](https://github.com/smallstep/agent/pull/1205) merged and shipped in 0.69.3, now on the stable channel, so the condition below is what current deb/rpm units and the NixOS module both report.

## Why

The NixOS section told readers *not* to manage `agent.yaml` with `environment.etc`, because "that produces a read-only symlink into the Nix store, and the service refuses to start." The diagnosis was wrong, and the advice ruled out the deployment NixOS users actually want — a customer raised exactly this while packaging the module for nixpkgs ([NixOS/nixpkgs#555971](https://github.com/NixOS/nixpkgs/pull/555971)).

Neither the symlink nor the file mode was the problem. The unit gated on `ConditionPathIsReadWrite=`, which per `systemd.unit(5)` tests whether the *underlying filesystem* is mounted read-only — not whether the file is writable — and it resolves symlinks. NixOS remounts `/nix/store` read-only at boot, so the check landed there and skipped the unit. The agent itself reads a mode 0444 `agent.yaml` without complaint and never writes to it; everything it writes lives in `/var/lib/step-agent`.

agent#1205 changes that gate to `ConditionPathExists=`, the question the unit meant to ask all along.

## Changes

- `platform/smallstep-agent.mdx` — no longer touched. The declarative example was dropped after review; #552 documents it as `services.step-agent.settings`.
- `platform/troubleshooting-agent.mdx` — update the "unmet condition check" callout to the new directive name, and note that older agents report the same check under the old one, so the message matches whichever version a reader has installed.

## Checks

`vale` run against the file before and after: no new alerts.

https://claude.ai/code/session_01RkEcowtxSeqvb2pLWC3u5J

### DIFF
--- a/platform/smallstep-agent.mdx
+++ b/platform/smallstep-agent.mdx
@@ -337,9 +337,24 @@ so a host with no <code>/dev/tpmrm0</code> cannot enroll yet.
     ```
 
     Registration writes `agent.yaml` into `/etc/step-agent`,
-    which systemd creates and keeps writable through `ConfigurationDirectory=`.
-    Do not manage `agent.yaml` with `environment.etc`:
-    that produces a read-only symlink into the Nix store, and the service refuses to start.
+    which systemd creates through `ConfigurationDirectory=`.
+
+    You can declare `agent.yaml` instead of registering interactively,
+    which is what makes a hands-off NixOS deployment possible.
+    Follow [Pre-registration via API](#pre-registration-via-api) to add and approve your devices
+    and to find your team slug and agent CA fingerprint, then declare the file and skip this step:
+
+    ```nix
+    environment.etc."step-agent/agent.yaml".text = ''
+      team: "[team name]"
+      fingerprint: "[agents CA fingerprint]"
+    '';
+    ```
+
+    The agent only ever reads this file.
+    Everything it writes lives in `/var/lib/step-agent`,
+    so serving `agent.yaml` from the Nix store is fine.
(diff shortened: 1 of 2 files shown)


platform/smallstep-agent.mdx
```diff
@@ -337,9 +337,24 @@ so a host with no <code>/dev/tpmrm0</code> cannot enroll yet.
...
+    You can declare `agent.yaml` instead of registering interactively,
+    which is what makes a hands-off NixOS deployment possible.
+    Follow [Pre-registration via API](#pre-registration-via-api) to add and approve your devices
+    and to find your team slug and agent CA fingerprint, then declare the file and skip this step:
+
+    ```nix
+    environment.etc."step-agent/agent.yaml".text = ''
+      team: "[team name]"
+      fingerprint: "[agents CA fingerprint]"
+    '';
```
Srylax, line 351: The module exposes services.step-agent.settings so instead of environment.etc this should be used:
```
services.step-agent.settings = {
 team = "[team name]";
 fingerprint = "[agents CA fingerprint]";
};
```

Srylax (changes requested):

PR #552, opened 2026-09-03
NixOS: declare the agent with services.step-agent; document edge releases

tashian:
Fixes [EFF-696](https://linear.app/smallstep/issue/EFF-696/update-install-docs-showing-how-to-install-an-edge-release).
Refs [EFF-698](https://linear.app/smallstep/issue/EFF-698/simplify-nixos-stable-install-docs-based-on-kylians-input): the stable-install rewrite is that issue's scope; it closes when the docs import the module from nixpkgs.

Rewrites the NixOS install section of the Smallstep Agent page and adds edge-release instructions. Replaces the `smallstep-agent.mdx` half of #549 (its `environment.etc` paragraph is subsumed by `settings`). #549 carries the `troubleshooting-agent.mdx` wording; agent#1205 shipped in 0.69.3, so this PR no longer touches that file.

## What changed

- **Install is one `services.step-agent` block.** Import the module, set `enable = true`, and declare `team` and `fingerprint` in `settings`. `nixos-rebuild switch` installs the package, starts the agent, and the device enrolls on first start once it has been added via the API. The interactive `step-agent register` step is now the alternative for empty `settings`, not the default.
- **Edge releases** subsection: point `services.step-agent.package` at an `overrideAttrs` of `pkgs.step-agent` with the edge tarball URL on `packages.smallstep.com` and the `sha256_sri` hash from the release manifest.
- Names the error a TPM-less host hits and updates the uninstall step.

The `services.step-agent` module with `enable`/`settings` ships from `files.smallstep.com/step-agent.nix` (identical to `extra/step-agent.nix` at the agent v0.69.3 tag) and has the same option surface as the module proposed upstream in [NixOS/nixpkgs#555971](https://github.com/NixOS/nixpkgs/pull/555971), so the `imports` line is the only thing that changes once that lands.

The `--skip-config` note exists because `register` silently drops a failed config write; that is filed as [OFF-21](https://linear.app/smallstep/issue/OFF-21/step-agent-register-config-write-failure-is-silently-discarded).

## Verified

Followed the new section on a nixos-unstable aarch64 VM: install, all units enabled, `agent.yaml` rendered from `settings`, `step-agent version` output as shown, and the edge override switching the running agent to 0.69.2. The edge snippet now shows 0.69.3, the stable release nixpkgs does not have yet; its hash was recomputed from the downloaded `step-agent_0.69.3_linux_amd64.tar.gz` and matches the manifest's `sha256_sri`.

The module change has shipped to `files.smallstep.com`, so step 3's `enable`/`settings` options exist for readers now.

### DIFF
--- a/platform/smallstep-agent.mdx
+++ b/platform/smallstep-agent.mdx
@@ -346,6 +346,9 @@ so on a host with no <code>/dev/tpmrm0</code> the service exits with <code>flag 
     To register interactively instead, leave `settings` out
     and run `sudo step-agent register [team name]` after the rebuild.
     The agent starts as soon as that writes `agent.yaml`.
+    With `settings` declared, you can still register a device interactively:
+    run `sudo step-agent register [team name] --skip-config`,
+    which registers the device without trying to rewrite the file you declared.
 
 4. Rebuild your system:
 
--- a/platform/smallstep-agent.mdx
+++ b/platform/smallstep-agent.mdx
@@ -341,7 +341,9 @@ so on a host with no <code>/dev/tpmrm0</code> the service exits with <code>flag 
     This declares the `step-agent` system user, the systemd service and its PKCS#11 socket, the `polkit` rules the agent needs, and the `p11-kit` module that publishes the agent's PKCS#11 server.
     `settings` is written to `/etc/step-agent/agent.yaml`.
     Every host in a fleet gets the same two values; nothing in the file is per-device.
-    The agent only reads it, and everything it writes lives in `/var/lib/step-agent`.
+    The agent uses the standard systemd directories:
+    it reads its config from `/etc/step-agent`, keeps its state in `/var/lib/step-agent`, and puts its sockets in `/run/step-agent`.
+    The daemon never writes to the config directory; only `step-agent register` does.
 
     To register interactively instead, leave `settings` out
     and run `sudo step-agent register [team name]` after the rebuild.
--- a/platform/smallstep-agent.mdx
+++ b/platform/smallstep-agent.mdx
@@ -390,14 +390,14 @@ services.step-agent.package = pkgs.step-agent.overrideAttrs (final: prev: rec {
   version = "0.69.2";
   src = pkgs.fetchurl {
     url = "https://packages.smallstep.com/edge/step-agent/linux/${version}/step-agent_${version}_linux_amd64.tar.gz";
-    sha256 = "1c8217188733c3a6ea558d399936825b426164d6e01d6f1c0ca4a67944f349fd";
+    hash = "sha256-HIIXGIczw6bqVY05mTaCW0JhZNbgHW8cDKSmeUTzSf0=";
   };
 });
 ```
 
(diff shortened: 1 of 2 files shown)


platform/smallstep-agent.mdx
```diff
@@ -317,29 +324,39 @@ so a host with no <code>/dev/tpmrm0</code> cannot enroll yet.
...
+      settings = {
+        team = "[team name]";
+        fingerprint = "[agent CA fingerprint]";
+      };
+    };
     ```
 
-3. Rebuild your system:
+    This declares the `step-agent` system user, the systemd service and its PKCS#11 socket, the `polkit` rules the agent needs, and the `p11-kit` module that publishes the agent's PKCS#11 server.
+    `settings` is written to `/etc/step-agent/agent.yaml`.
```
Srylax, line 342: that is actually incorrect. It does not write it into agent.yaml but rather into the nix store and refernces it via --config

platform/smallstep-agent.mdx
```diff
@@ -317,29 +324,39 @@ so a host with no <code>/dev/tpmrm0</code> cannot enroll yet.
...
+    Every host in a fleet gets the same two values; nothing in the file is per-device.
+    The agent only reads it, and everything it writes lives in `/var/lib/step-agent`.
```
Srylax: and `/run/step-agent/` and `/etc/step-agent/` :)))

platform/smallstep-agent.mdx
```diff
@@ -350,10 +367,37 @@ so a host with no <code>/dev/tpmrm0</code> cannot enroll yet.
...
+nixpkgs carries a stable release, and only once it has been packaged there.
+To run an edge release, or a stable release your channel does not have yet,
+point `services.step-agent.package` at the release tarball on `packages.smallstep.com`:
+
+```nix
+services.step-agent.package = pkgs.step-agent.overrideAttrs (final: prev: rec {
+  version = "0.69.2";
+  src = pkgs.fetchurl {
+    url = "https://packages.smallstep.com/edge/step-agent/linux/${version}/step-agent_${version}_linux_amd64.tar.gz";
+    sha256 = "1c8217188733c3a6ea558d399936825b426164d6e01d6f1c0ca4a67944f349fd";
```
Srylax: Is it wise to use the hex-form? What happens if it gets a mismatch? will nix print the updated hex form or the updated SNI? could be confusing

Srylax (changes requested):

platform/smallstep-agent.mdx
tashian: Good point, I just tested this and it normalizes everything to SRI form. I will add a field and publish the SRI to releases.smallstep.com metadata.

tashian: I've added a sha256_sri field to our JSON manifest generator (and backfilled past releases). 
See https://packages.smallstep.com/edge/step-agent/linux/0.69.2/index.json for example.

platform/smallstep-agent.mdx
tashian: Thanks, it names all three directories now.

darkfronza (approved):

